PR #3, opened 2019-02-15
Factor out drawing context from platform window implementation

drobilla:


drobilla:
@x42 There, find something better to complain about :P

x42:
Thank you!

PR #7, opened 2019-04-17
Prevent infinite loop in puglProcessEvents() (OSX)

7890:
Please see commit message

### DIFF
--- a/pugl/pugl_osx.m
+++ b/pugl/pugl_osx.m
@@ -686,17 +686,13 @@ puglWaitForEvent(PuglView* view)
 PuglStatus
 puglProcessEvents(PuglView* view)
 {
-	while (true) {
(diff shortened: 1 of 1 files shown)


drobilla:
Like the documentation of `puglProcessEvents` says, it needs to "Process all pending window events", just like on the other platforms.

7890:
I don't know why but from what I see this does a blocking wait (it stays there until a new event is created):
```
view->impl->nextEvent = [view->impl->window nextEventMatchingMask: NSAnyEventMask];
```
When done inside while(true) in puglProcessEvents() which again is wrapped in the main() application loop, it will never return since it waits for an event to happen and then the condition for continuing the loop is always true.
You can easily verify this by trying 'q' in pugl_test. It never gets past puglProcessEvents().

drobilla:
9734823

PR #9, opened 2019-06-23
Some fixes

swesterfeld:
SpectMorph is using pugl for its UI. Here are a few fixes that I've made that should be good enough for inclusion in upstream.

Event times: I use the event timestamp for double click detection (two clicks, less than 400ms apart). macOS uses a double timestamp for events in seconds, whereas the other platform use milliseconds as integers for the event time. To be able to use the same code for all platforms, I suggest always using double seconds as timestamps, this should be good enough to represent both.

Backspace/delete is not 0x8/0x7f on all platforms, so I suggest using the same strategy as for function keys: define PUGL_KEY_BACKSPACE | PUGL_KEY_DELETE, and map whatever platform specific code this has to these pugl keys.

On macOS current pugl has event->key.character and event->key.special set (i.e. if you use arrow keys), this is not consistent with the other platforms. So I've added a commit which enforces that a special key will always have character == 0.

The other fairly small fixes just fix bugs.

Ideally one day I would be able to use an unmodified version of pugl for SpectMorph, but there are a few more things to sort out. One main issue is that I need to resize the window if the user changes plugin zoom level, currently that is done using my own version of 
```
void puglSetResizeFunc(PuglView* view, PuglResizeFunc resizeFunc);
void puglPostResize(PuglView* view);
```
but here I currenly don't have a clean enough patch for upstream inclusion so I left this out at the moment.

### DIFF
--- a/pugl/pugl.h
+++ b/pugl/pugl.h
@@ -116,9 +116,7 @@ typedef enum {
    Convenience symbols for ASCII control characters.
 */
 typedef enum {
-	PUGL_CHAR_BACKSPACE = 0x08,
 	PUGL_CHAR_ESCAPE    = 0x1B,
-	PUGL_CHAR_DELETE    = 0x7F
 } PuglChar;
 
 /**
@@ -152,6 +150,8 @@ typedef enum {
 	PUGL_KEY_F10,
 	PUGL_KEY_F11,
 	PUGL_KEY_F12,
+        PUGL_KEY_BACKSPACE,
+        PUGL_KEY_DELETE,
 	PUGL_KEY_LEFT,
 	PUGL_KEY_UP,
 	PUGL_KEY_RIGHT,
--- a/pugl/pugl.h
+++ b/pugl/pugl.h
@@ -218,7 +218,7 @@ typedef struct {
 	PuglEventType type;        /**< PUGL_BUTTON_PRESS or PUGL_BUTTON_RELEASE. */
 	PuglView*     view;        /**< View that received this event. */
 	uint32_t      flags;       /**< Bitwise OR of PuglEventFlag values. */
-	uint32_t      time;        /**< Time in milliseconds. */
+	double        time;        /**< Time in seconds. */
 	double        x;           /**< View-relative X coordinate. */
 	double        y;           /**< View-relative Y coordinate. */
 	double        x_root;      /**< Root-relative X coordinate. */
@@ -288,7 +288,7 @@ typedef struct {
 	PuglEventType type;        /**< PUGL_KEY_PRESS or PUGL_KEY_RELEASE. */
 	PuglView*     view;        /**< View that received this event. */
 	uint32_t      flags;       /**< Bitwise OR of PuglEventFlag values. */
-	uint32_t      time;        /**< Time in milliseconds. */
+	double        time;        /**< Time in seconds. */
 	double        x;           /**< View-relative X coordinate. */
 	double        y;           /**< View-relative Y coordinate. */
 	double        x_root;      /**< Root-relative X coordinate. */
@@ -308,7 +308,7 @@ typedef struct {
 	PuglEventType    type;     /**< PUGL_ENTER_NOTIFY or PUGL_LEAVE_NOTIFY. */
 	PuglView*        view;     /**< View that received this event. */
 	uint32_t         flags;    /**< Bitwise OR of PuglEventFlag values. */
-	uint32_t         time;     /**< Time in milliseconds. */
+	double           time;     /**< Time in seconds. */
 	double           x;        /**< View-relative X coordinate. */
 	double           y;        /**< View-relative Y coordinate. */
 	double           x_root;   /**< Root-relative X coordinate. */
@@ -324,7 +324,7 @@ typedef struct {
 	PuglEventType type;        /**< PUGL_MOTION_NOTIFY. */
 	PuglView*     view;        /**< View that received this event. */
 	uint32_t      flags;       /**< Bitwise OR of PuglEventFlag values. */
-	uint32_t      time;        /**< Time in milliseconds. */
+	double        time;        /**< Time in seconds. */
 	double        x;           /**< View-relative X coordinate. */
 	double        y;           /**< View-relative Y coordinate. */
 	double        x_root;      /**< Root-relative X coordinate. */
@@ -347,7 +347,7 @@ typedef struct {
 	PuglEventType type;        /**< PUGL_SCROLL. */
 	PuglView*     view;        /**< View that received this event. */
 	uint32_t      flags;       /**< Bitwise OR of PuglEventFlag values. */
-	uint32_t      time;        /**< Time in milliseconds. */
+	double        time;        /**< Time in seconds. */
 	double        x;           /**< View-relative X coordinate. */
 	double        y;           /**< View-relative Y coordinate. */
 	double        x_root;      /**< Root-relative X coordinate. */
--- a/pugl/pugl_osx.m
+++ b/pugl/pugl_osx.m
@@ -695,7 +695,11 @@ puglProcessEvents(PuglView* view)
 		// Get the next event, or use the cached one from puglWaitForEvent
 		if (!view->impl->nextEvent) {
 			view->impl->nextEvent = [view->impl->window
-			                            nextEventMatchingMask: NSAnyEventMask];
+			                            nextEventMatchingMask: NSAnyEventMask
+                                                    untilDate:nil
+                                                    inMode:NSDefaultRunLoopMode
+                                                    dequeue:YES];
+
 		}
 
 		if (!view->impl->nextEvent) {
--- a/pugl/pugl_osx.m
+++ b/pugl/pugl_osx.m
@@ -595,8 +595,8 @@ puglCreateWindow(PuglView* view, const char* title)
 		[impl->glview setAutoresizingMask:NSViewNotSizable];
 	}
 
-	if (view->transient_parent) {
-		NSView* pview = (NSView*)view->transient_parent;
+	if (view->parent) {
+		NSView* pview = (NSView*)view->parent;
 		[pview addSubview:impl->glview];
 		[impl->glview setHidden:NO];
 	} else {
--- a/pugl/pugl_osx.m
+++ b/pugl/pugl_osx.m
@@ -439,7 +439,8 @@ keySymToSpecial(PuglView* view, NSEvent* ev)
 	const NSPoint      rloc  = [NSEvent mouseLocation];
 	const NSString*    chars = [event characters];
 	const char*        str   = [chars UTF8String];
-	const uint32_t     code  = puglDecodeUTF8((const uint8_t*)str);
+        const PuglKey      special = keySymToSpecial(puglview, event);
+        const uint32_t     code  = special ? 0 : puglDecodeUTF8((const uint8_t*)str);
 	PuglEventKey       ev    =  {
 		PUGL_KEY_PRESS,
 		puglview,
@@ -466,6 +467,8 @@ keySymToSpecial(PuglView* view, NSEvent* ev)
 	const NSPoint      rloc  = [NSEvent mouseLocation];
 	const NSString*    chars = [event characters];
 	const char*        str   = [chars UTF8String];
+        const PuglKey      special = keySymToSpecial(puglview, event);
+        const uint32_t     code  = special ? 0 : puglDecodeUTF8((const uint8_t*)str);
 	const PuglEventKey ev    =  {
 		PUGL_KEY_RELEASE,
 		puglview,
@@ -477,7 +480,7 @@ keySymToSpecial(PuglView* view, NSEvent* ev)
 		[[NSScreen mainScreen] frame].size.height - rloc.y,
 		getModifiers(puglview, event),
 		[event keyCode],
-		puglDecodeUTF8((const uint8_t*)str),
+		(code != 0xFFFD) ? code : 0,
 		keySymToSpecial(puglview, event),
 		{ 0, 0, 0, 0, 0, 0, 0, 0 },
 		false,
--- a/pugl/pugl_osx.m
+++ b/pugl/pugl_osx.m
@@ -264,6 +264,8 @@ keySymToSpecial(PuglView* view, NSEvent* ev)
 		case NSF10FunctionKey:        return PUGL_KEY_F10;
 		case NSF11FunctionKey:        return PUGL_KEY_F11;
 		case NSF12FunctionKey:        return PUGL_KEY_F12;
+                case NSDeleteCharacter:       return PUGL_KEY_BACKSPACE;
+                case NSDeleteFunctionKey:     return PUGL_KEY_DELETE;
 		case NSLeftArrowFunctionKey:  return PUGL_KEY_LEFT;
 		case NSUpArrowFunctionKey:    return PUGL_KEY_UP;
 		case NSRightArrowFunctionKey: return PUGL_KEY_RIGHT;
--- a/pugl/pugl_win.cpp
+++ b/pugl/pugl_win.cpp
@@ -250,6 +250,8 @@ keySymToSpecial(int sym)
 	case VK_F10:     return PUGL_KEY_F10;
 	case VK_F11:     return PUGL_KEY_F11;
 	case VK_F12:     return PUGL_KEY_F12;
+        case VK_BACK:    return PUGL_KEY_BACKSPACE;
+        case VK_DELETE:  return PUGL_KEY_DELETE;
 	case VK_LEFT:    return PUGL_KEY_LEFT;
 	case VK_UP:      return PUGL_KEY_UP;
 	case VK_RIGHT:   return PUGL_KEY_RIGHT;
--- a/pugl/pugl_win.cpp
+++ b/pugl/pugl_win.cpp
@@ -280,6 +280,12 @@ getModifiers()
 	return mods;
 }
 
+static double
+translateMessageTime()
+{
+  return GetMessageTime() / 1000.; /* translate ms to seconds */
+}
+
 static void
 initMouseEvent(PuglEvent* event,
                PuglView*  view,
@@ -296,7 +302,7 @@ initMouseEvent(PuglEvent* event,
 		ReleaseCapture();
 	}
 
-	event->button.time   = GetMessageTime();
+	event->button.time   = translateMessageTime();
 	event->button.type   = press ? PUGL_BUTTON_PRESS : PUGL_BUTTON_RELEASE;
 	event->button.x      = GET_X_LPARAM(lParam);
 	event->button.y      = GET_Y_LPARAM(lParam);
@@ -312,7 +318,7 @@ initScrollEvent(PuglEvent* event, PuglView* view, LPARAM lParam, WPARAM wParam)
 	POINT pt = { GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam) };
 	ScreenToClient(view->impl->hwnd, &pt);
 
-	event->scroll.time   = GetMessageTime();
+	event->scroll.time   = translateMessageTime();
 	event->scroll.type   = PUGL_SCROLL;
 	event->scroll.x      = pt.x;
 	event->scroll.y      = pt.y;
@@ -357,7 +363,7 @@ initKeyEvent(PuglEvent* event, PuglView* view, bool press, LPARAM lParam)
 	ScreenToClient(view->impl->hwnd, &rpos);
 
 	event->key.type      = press ? PUGL_KEY_PRESS : PUGL_KEY_RELEASE;
-	event->key.time      = GetMessageTime();
+	event->key.time      = translateMessageTime();
 	event->key.state     = getModifiers();
 	event->key.x_root    = rpos.x;
 	event->key.y_root    = rpos.y;
@@ -499,7 +505,7 @@ handleMessage(PuglView* view, UINT message, WPARAM wParam, LPARAM lParam)
 		ClientToScreen(view->impl->hwnd, &pt);
 
 		event.motion.type    = PUGL_MOTION_NOTIFY;
-		event.motion.time    = GetMessageTime();
+		event.motion.time    = translateMessageTime();
 		event.motion.x       = GET_X_LPARAM(lParam);
 		event.motion.y       = GET_Y_LPARAM(lParam);
 		event.motion.x_root  = pt.x;
--- a/pugl/pugl_win.cpp
+++ b/pugl/pugl_win.cpp
@@ -137,7 +137,7 @@ puglCreateWindow(PuglView* view, const char* title)
 		return NULL;
 	}
 
-	int winFlags = WS_POPUPWINDOW | WS_CAPTION;
+	int winFlags = view->parent ? WS_CHILD : WS_POPUPWINDOW | WS_CAPTION;
 	if (view->hints.resizable) {
 		winFlags |= WS_SIZEBOX;
 		if (view->min_width || view->min_height) {
--- a/pugl/pugl_x11.c
+++ b/pugl/pugl_x11.c
@@ -297,6 +297,7 @@ translateEvent(PuglView* view, XEvent xevent)
 		event.any.flags |= PUGL_IS_SEND_EVENT;
 	}
 
+        const double time_ms_to_s = 1 / 1000.;
 	switch (xevent.type) {
 	case ClientMessage: {
 		char* type = XGetAtomName(view->impl->display,
@@ -323,7 +324,7 @@ translateEvent(PuglView* view, XEvent xevent)
 		break;
 	case MotionNotify:
 		event.type           = PUGL_MOTION_NOTIFY;
-		event.motion.time    = xevent.xmotion.time;
+		event.motion.time    = xevent.xmotion.time * time_ms_to_s;
 		event.motion.x       = xevent.xmotion.x;
 		event.motion.y       = xevent.xmotion.y;
 		event.motion.x_root  = xevent.xmotion.x_root;
@@ -334,7 +335,7 @@ translateEvent(PuglView* view, XEvent xevent)
 	case ButtonPress:
 		if (xevent.xbutton.button >= 4 && xevent.xbutton.button <= 7) {
 			event.type           = PUGL_SCROLL;
-			event.scroll.time    = xevent.xbutton.time;
+			event.scroll.time    = xevent.xbutton.time * time_ms_to_s;
 			event.scroll.x       = xevent.xbutton.x;
 			event.scroll.y       = xevent.xbutton.y;
 			event.scroll.x_root  = xevent.xbutton.x_root;
@@ -356,7 +357,7 @@ translateEvent(PuglView* view, XEvent xevent)
 			event.button.type   = ((xevent.type == ButtonPress)
 			                       ? PUGL_BUTTON_PRESS
 			                       : PUGL_BUTTON_RELEASE);
-			event.button.time   = xevent.xbutton.time;
+			event.button.time   = xevent.xbutton.time * time_ms_to_s;
 			event.button.x      = xevent.xbutton.x;
 			event.button.y      = xevent.xbutton.y;
 			event.button.x_root = xevent.xbutton.x_root;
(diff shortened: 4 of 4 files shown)


pugl/pugl.h
```diff
@@ -218,7 +218,7 @@ typedef struct {
 	PuglEventType type;        /**< PUGL_BUTTON_PRESS or PUGL_BUTTON_RELEASE. */
 	PuglView*     view;        /**< View that received this event. */
 	uint32_t      flags;       /**< Bitwise OR of PuglEventFlag values. */
-	uint32_t      time;        /**< Time in milliseconds. */
+	double        time;        /**< Time in seconds. */
```
drobilla: This gives me pause... I agree that seconds is the best unit in general, but worry a bit about the imprecision of floating point.  Could not being able to get back to the system-provided integer value possibly be a problem?

I suppose pugl code should stay in "pugl land", so probably not.

pugl/pugl.h
```diff
@@ -152,6 +150,8 @@ typedef enum {
 	PUGL_KEY_F10,
 	PUGL_KEY_F11,
 	PUGL_KEY_F12,
+        PUGL_KEY_BACKSPACE,
```
drobilla: I am not sure this is a good idea.  It's a bit of a slippery slope.  They _are_ ASCII characters, which is the rule used to make this distinction in the API.  Shouldn't backends just deal with this if necessary?

pugl/pugl_osx.m
```diff
@@ -439,7 +439,8 @@ - (void) keyDown:(NSEvent*)event
 	const NSPoint      rloc  = [NSEvent mouseLocation];
 	const NSString*    chars = [event characters];
 	const char*        str   = [chars UTF8String];
-	const uint32_t     code  = puglDecodeUTF8((const uint8_t*)str);
+        const PuglKey      special = keySymToSpecial(puglview, event);
+        const uint32_t     code  = special ? 0 : puglDecodeUTF8((const uint8_t*)str);
```
drobilla: Interesting.  What ended up coming out here, out of curiosity?

pugl/pugl_osx.m
```diff
@@ -695,7 +695,11 @@ - (void) flagsChanged:(NSEvent*)event
 		// Get the next event, or use the cached one from puglWaitForEvent
 		if (!view->impl->nextEvent) {
 			view->impl->nextEvent = [view->impl->window
-			                            nextEventMatchingMask: NSAnyEventMask];
+			                            nextEventMatchingMask: NSAnyEventMask
+                                                    untilDate:nil
+                                                    inMode:NSDefaultRunLoopMode
```
drobilla: Ah, this is (hopefully) the magic incantation that makes this work...

pugl/pugl.h
swesterfeld: I can't think of a case where you would want to get the system integer back. At least the way I use the timestamps, what matters is the difference of two timestamps, and if we alllow doubles, then fractional seconds are the more natural representation.

If you do want the system provided integer back, you can multiply with 1000 and round to the nearest integer value. Since doubles have 52 bits for representing the "number part", this should be perfectly accurate for 32bit integer timestamps. For 64bit integer timestamps, it should still be accurate if the timestamp is not too large (assuming it initially starts at zero it should never be too large).

pugl/pugl.h
swesterfeld: These are ASCII characters. But what I want to know if I use pugl is what key was pressed by the user. And unfortunately if the user presses backspace, I get 8 on X11 and Windows and 0x7f on macOS. For delete I get 0x7f on X11, 0x2e on Windows and 0xf728 on macOS. The situation is similar to what happens for arrow keys or function keys. These keys have platform dependent codes, so pugl - as a portability layer - should interpret and map this for me.

pugl/pugl_osx.m
swesterfeld: This was created while I implemented a LineEdit widget. The LineEdit widget first checks if a character is to be entered by checking (code != 0). If yes, the character is added to the text. If no character was entered it also checks for arrow keys.

What was happening on macOS is that if you use arrow keys, then the code of the keyboard event was set to the value for the arrow keys, and code was != 0. Instead of having arrow keys move the cursor, the character was added to the text inside the line edit. So this was the reason for setting code for all special keys to zero.

An alternative implementation would be to check if code is in the reserved unicode section (U+E000–U+F8FF) which seems to be where all macOS special key codes are, and unset code for these events.

pugl/pugl.h
drobilla: Strictly speaking raw keycode is the only thing that can reliably be used to know "what key was pressed by the user", but I agree these should be consistent wherever possible.

My concern is that pugl deliberately re-uses UTF-8 to avoid having thousands and thousands of lines of key symbols like in X and similar.  So, the current rule is, PuglKey is for things that do not have an ASCII code.  After this change, the rule is..... well... it's a PuglKey if some platform specific crap made it convenient.  I would like to stabilize the API as soon as possible, so this makes me nervous.  Some platform backend could come along some day that happens to produce something else for tab (or whatever), and more and more things have to move, which (semantically) breaks the API and never ends.

Sorry if this seems pedantic, but this same thing probably applies to all sorts of control characters that might be on a keyboard somewhere, and I'd like to constrain this stuff as much as possible.  Ideally PuglKey wouldn't even exist at all, but unfortunately Unicode doesn't cover everything.

So, the question is: is there any reason the backend just translate them where necessary to preserve the public API as it was and avoid inventing new codepoints for things that already have them?

pugl/pugl_osx.m
drobilla: Seems like special and code setting needs to be factored into the same function so this can all be handled correctly in one place?

pugl/pugl.h
drobilla: Fair enough.  This struct is a blatant rip off of the X one (which means it also almost exactly matches the Gtk one), but I can't think of a particularly good reason not to do this here either.

drobilla:
I need to fire up my other machines and do some manual testing to wrap my head around this (oh, for a way to automatically test this stuff...), but I'll try to get this merged this week sometime.

It would be awesome to get you (and everyone else!) on unmodified upstream pugl.  The "copy paste code reuse" idea was a big mistake.  I am going to include pugl in the upcoming "lv2kit" which should help provide some additional incentive in that direction...

drobilla:
For prosperity (rebasing): f3782ca 0962252 cbc869f ca87f40

drobilla:
For prosperity (rebasing): f3782ca 0962252 cbc869f ca87f40

pugl/pugl_osx.m
swesterfeld: Right. I'd still have to try implementing it to see if its more elegant, but it could be. So whats the strategy now during the merge? Should I commit on top of this branch, or better wait until you merged stuff so far?

pugl/pugl.h
swesterfeld: The case that control keys are already in ASCII is rare. Maybe 
- backspace
- escape
- tab
- delete
- return/enter

So we're not speaking about an ever-growing list of additions if we consistently use PuglKey also for control keys that already have an ASCII value. Rather this is a fixed set of a few ASCII characters that pugl would map to a new code point. From an API point of view, I'd prefer PuglKey handling all special keys. The alternative would be some keys have a code (PUGL_CHAR_BACKSPACE) and others are special (PUGL_KEY_F1). So yes, going this way we may need to add a few more keys (like PUGL_KEY_TAB) for consistency reasons, but then we should have a solution that is sane, maintainable and stable.

Let me add that for API design issues, it is good to look at existing solutions, and for instance Qt5 also has an enum, which contains keys like Qt::Key_Backspace or Qt::Key_Escape and Qt::Key_F1 https://doc.qt.io/qt-5/qt.html#Key-enum

swesterfeld: > So, the question is: is there any reason the backend just translate them where necessary to preserve the public API as it was and avoid inventing new codepoints for things that already have them?

As already stated I like the PUGL_KEY_* approach better.  But: technically it would be possible to map whatever native delete/backspace code there is to code==8 and code==0x7f which would have maximum compatibility with your existing API.

drd-ableton: Yes, but Qt and many/most similar libs have more lines of just these declarations than pugl has code, period :)  This is exactly the thing I want to avoid.

That said, you've almost convinced me.  It is true that there is only so much stuff in ASCII...

Really it's about the simplest and most minimal thing, which makes me wonder, maybe we should not only define these (as you have), but the API should simply exploit the fact that these are all in the Unicode "namespace" and just remove the `special` field entirely?

drobilla: Yes, but Qt and many/most similar libs have more lines of just these declarations than pugl has code, period :)  This is exactly the thing I want to avoid.

That said, you've almost convinced me.  It is true that there is only so much stuff in ASCII...

Really it's about the simplest and most minimal thing, which makes me wonder, maybe we should not only define these (as you have), but the API should simply exploit the fact that these are all in the Unicode "namespace" and just remove the `special` field entirely?

pugl/pugl_osx.m
drobilla: As you like.  I can never figure out this PR workflow, ideally we'd just work on the same branch...

I can rebase anything, though, so whatever.  I'll try to pop some stuff off and get them on master.  I did some testing on MacOS and it seems fine, I'll try Windows now.

drobilla:
I merged 3 of your commits, and a whole bunch of other fixes/improvements I found along the way, to master.  https://github.com/drobilla/pugl/tree/some-fixes is now everything left, rebased on that.

No idea if the embedding works, though... I should really write test programs for that one of these days.

pugl/pugl.h
drobilla: Like so: https://github.com/drobilla/pugl/tree/uniform-keys

(Only implemented for X11 to try it out, but I think the API is much less confusing this way?)

swesterfeld: I think removing `special` is a good idea. Your uniform-keys branch code looks ok to me. However then we need something that can be used to check whether this is an ordinary character (which my LineEdit widget would add to its text) or a special character (which it wouldn't add). Something like puglKeyIsSpecial (character) which would check for 0xe000..0xf8ff. Or enum values: PUGL_KEY_SPECIAL_FIRST ... PUGL_KEY_SPECIAL_LAST. Or a boolean in the event structure.

By the way, I've thought about the backspace problem some more, and I think there is a very good reason to prefer PUGL_KEY_BACKSPACE (0xe000+something) over PUGL_CHAR_BACKSPACE (8), even if we remove `special` completely. The problem is, that pugl currently maps two keyboard events to the same code point: backspace and Ctrl+H. So what happens is that if I use a LineEdit widget with your pugl, I can effectively delete characters using backspace (which is what I want) and Ctrl+H (which is not what I want). Also if I define a shortcut that is bound to Ctrl+H, I can trigger it with backspace.

With my changes PUGL_KEY_BACKSPACE is a special character, somewhere above 0xe000 both problems do not arise. The LineEdit reacts to backspace and backspace only, and a Ctrl+H shortcut can no longer be triggered by backspace.

There are pathologic workarounds like checking the modifiers and if Ctrl is not held while a Ctrl+H is received then it is probably backspace, but I'd prefer having pugl not map this to the same character in the first place. Of course other special keys should not be represented by Ctrl+something ASCII for this same reason (so we need something new for tab, return and enter).

drobilla: Hm.  Good point, though it seems that means that we haven't quite figured this one out yet and moving backspace (etc) to the reserved range doesn't fix this.  For example, if you want that property of ASCII to go away and actually get ^H, the problem (on X11 anyway) is not really that backspace sends backspace, but that ^H sends backspace (not control and then H).  Similarly ^J sends a newline, and so on.

Maybe we can just clear the mods of the XEvent before using X*LookupString...

drobilla: That's df647f4

I feel like shift is weird now.  If this is representing _key_ then it probably shouldn't be shifted...

drobilla:
So I did quite a bit of tinkering with the key stuff on https://github.com/drobilla/pugl/tree/uniform-keys

In the process I managed to get dead keys working on Windows, which is nice (it was totally broken before), but I haven't done MacOS yet.  I'm kind of debating whether to emulate the API of Windows more and just provide two completely separate events for key press/release and text input so there's no risk of confusing the two.  It's more or less impossible to "merge" them on Windows, so there's two events anyway, so maybe they should just have different types altogether...

The "unshifted unicode point" idea seems to be solid, but I'm not sure about the rest of the details.  I haven't actually implemented a text entry widget with pugl, so I'm kind of just guessing at what's convenient.  The shifted one seems kind of useless for identifying keys, so maybe it should just go, but this is another argument for separate events (`PuglEventText` would be shifted, to match the string).  Thoughts?

swesterfeld:
First of all, there seems to be a small bug in your uniform-keys branch on X11: if I hit F1, I'd expect unicode 0xe000, but I get unicode 0x0000 (similar for other keys). It seems that the `lookupString` return code is not greater than 0 in here:

```if (lookupString(view->impl->xic, xevent, event->string, &sym) > 0)```

As for shortcuts with your new event key normalization: Ctrl/Shift/Alt+A..Z should now work properly (I used to normalize these manually for old pugl), also Ctrl/Shift/Alt+0..9. Just check unshifted key and compare modifiers and that's it. Also the collision problem for backspace/tab/... is really solved this way.

I can however think of one class of keyboard shortcuts which I don't know how to handle with your API by checking key and comparing modifiers: lets look at for example `Ctrl+;`.

On a us keyboard layout, there is a key for `;`, so pugl_test shows for `Ctrl+;`:

```Text        code 47 char U+003B key U+003B (;)```

On a de (german) keyboard layout, I have to use `Shift+,` to get a `;`, and therefore pugl_test prints for `Ctrl+;` aka `Ctrl+Shift+,`

```Text        code 59 char U+003B key U+002C (;)```

As you see, key is normalized to `;` (us) or `,` (de) depending on keyboard layout, so checking for `key==';'` will not reliably detect `Ctrl+;`. Also a 1:1 comparision of modifiers is not possible, since it would be `Ctrl` (us) or `Ctrl+Shift` (de). So what this means is that for such shortcuts I don't really know how to use the keyboard event as generated by pugl to figure out if that shortcut was used. On the other hand it may be acceptable to just avoid defining such shortcuts in pugl UIs. In this case I'd agree that there is no reason to keep shifted character, and we should just have unshifted key and that is all.

Finally a few thoughts on what is required for the LineEdit widget: it is somewhat convenient to be able to handle all editing operations (such as delete or entering an 'A') in one single function, because you can add common code (like emitting a signal if the text was changed) in one place at the end of this editing function. So this kind of suggests both should be in one key press event handler. However this is not a hard requirement, it would probably be not so bad to refactor this to two different functions so I really can't say whether using one event type (key press) or two event types (key press and text) is better.

When new text is entered, I never even look at the keyevent character, key or modifiers. So I don't care if they are shifted, unshifted, normalized or even zero. I just use the utf8 string the key press event contains and update the string (see `key_press_event` function https://github.com/swesterfeld/spectmorph/blob/master/glui/smlineedit.hh). So as long as I can reliably detect if the event contains text (then I need to add it to the text of the widget) or not (to reliably detect if something is a control character like "left arrow") everything is fine.

**Comment by drobilla** [comment]
I think shortcuts like that are going to be problematic in the same way regardless of API.  There's really only two things any hypothetical API can tell you about this stuff: which key was pressed, and which text (if any) that corresponds to.  Ctrl + (something that isn't a key) is always going to be different than ctrl+key somehow, the keypresses and modifier state are genuinely different.

That said, it seems you can implement it easily enough if you really want to.  You have the ';' both in the text, and in the shifted charcter code (U+003B), and you have the fact that ctrl is pressed.  Information-wise everything is available, except for "shifted with shift but not with ctrl".  That can be dealt with easily either in client code or in pugl though (I'm not sure, but prefer the simple modified/unmodified thing vs treating different ones specially in pugl which seems like it will just cause more problems in the future because it's not what somebody is going to want for something).

**Comment by swesterfeld** [comment]
> only two things any hypothetical API can tell you

I disagree. For instance, of course Qt5 supports defining a shortcut for `Ctrl+;`. I just create a QShortcut for `Ctrl+;` and that is it. I don't have to guess, and it works with any keyboard layout, Qt5 does this for me. Of course QShortcut belongs rather into the SpectMorph toolkit than into pugl, as it is high level API.

Now I investigated what low level API QShortcut is using to detect `Ctrl+;`. There is a private part of the Qt API that defines QKeyMapper, this is platform specific. And this QKeyMapper can give you a list of possible keys for a QKeyEvent*. So detecting `Ctrl+;` with Qt5 private API can be written as:
```
void
keyPressEvent (QKeyEvent *event)
{
  for (auto k : QKeyMapper::possibleKeys (event))
    {
      if (k == ';' + Qt::CTRL)
        printf ("Ctrl+; detected.\n");
    }
}
```
For the record QKeyMapper::possibleKeys returns only `Ctrl+;` as possible input for us layout, whereas it returns `Ctrl+Shift+,` and `Ctrl+;` for de layout.

**Comment by swesterfeld** [comment]
I've meanwhile looked at the platform specific `possibleKeys (QKeyEvent *event)` code for XCB, Windows and macOS in Qt. To do their work, they use

- `event->modifiers()`
- `event->nativeVirtualKey()`
- `event->nativeModifiers()`
- `event->nativeScanCode()`

This seems like it doesn't really use what pugl calls `character` or `key` but rather operates on the platform specific part of the event. Not sure if `keycode`and `state` would be sufficient for pugl. If yes, and if we wish to avoid returning a list of things (bad for C API), it should be possible to implement a

```
bool puglKeyCompare (PuglKeyEvent *key_event, uint32_t state, uint32_t character);
```
which could then be used as
```
  if (puglKeyCompare (key_event, PUGL_MOD_CTRL, ';'))
    printf ("Ctrl+; detected.\n");
```
so it would basically check if the shortcut `state` and `character` was in the `possibleKeys` list of `key_event`. It is still a bit of work which has to be repeated for each platform, and it now looks like it doesn't really affect the API freeze for the keyboard API we were discussing earlier. So.... if you feel like the shortcut matching should be postponed, then ok, we have other issues which I also think have higher priority. Like getting backspace and delete to work.

Just one example implementation of possibleKeys for XCB:
https://github.com/qt/qtbase/blob/11d9af5d957ac391a0c6cfdea4819df44b2b1560/src/plugins/platforms/xcb/qxcbkeyboard.cpp#L1013

**Comment by drobilla** [comment]
I don't think it would be very good to have the core API of Pugl use only scancodes (and so be useless for most applications) and require a higher level API written from the ground up to get text.  As you say, that's more of a toolkit thing.

Anyway, I managed to get everything working as far as I can tell on all three platforms, with dead keys and composing and all the rest of the tricky bits: https://github.com/drobilla/pugl/commit/b11f5ff754df4e189792c18521e758ecb388a648

**Comment by drobilla** [comment]
I merged the double/seconds change, and more or less completely rewrote the keyboard/text input.  I've tested it quite a bit on all three platforms, and it seems to work with all the tricky situations like compose and dead keys.  There's probably still some edge cases that aren't right (and some things that just don't work at all, like CapsLock on Cocoa), but it's dramatically better than it was anyway.

I fixed a bunch of other stuff as well, let me know if you have any issues.

**Comment by swesterfeld** [comment]
> let me know if you have any issues.

I did some tests today, compiling my SpectMorph plugin against your pugl. Text events work nicely. Here is a preliminary results

macOS
- backspace and delete don't work
- the y - coordinates of mouse events are wrong (if I make - (BOOL) isFlipped return NO, things work)
- I can no longer build with 10.9 SDK - currently I target macOS >= 10.9 for SpectMorph

Of course the last point may not be much of an issue for pugl upstream, you can require whatever you think is appropriate. It used to build against 10.9 SDK before, though. I probably wouldn't update my build environment just for pugl, though.

Windows
- delete doesn't work
- non-standard location for cairo doesn't work, should replace in cairo_gl.h
```#include <cairo/cairo.h>```
with
```#include <cairo.h>```
- I saw some redrawing glitches on windows, but haven't figured out why it happens

Linux
- initial redraw of a window doesn't work - after that it seems to work - I haven't figured out the cause yet

general
- struct PuglEventText time field should probably be double
- you didn't merge the special key translation for backspace and delete - I believe you can do it unmodified from https://github.com/swesterfeld/pugl/commit/0361ac57fba1d55e00dba3f55ff610979e0cbe68 which should fix these keys on windows/macOS

**Comment by drobilla** [comment]
Sigh.  It builds here with `-mmacosx-version-min` going all the way back to 10.6.  Can you be more specific?

**Comment by drobilla** [comment]
Initial display on X11 works for me, this could be a window manager specific thing.

Everything else should be fixed, thanks.  As you can see I suck at manual testing :)

**Comment by drobilla** [comment]
Blech.  I guess in the context of pugl I can't complain about people breaking APIs, but still...

**Comment by drobilla** [comment]
2e3e935
3359c78

**Comment by swesterfeld** [comment]
Ok, some progress on

> Linux: initial redraw of a window doesn't work - after that it seems to work - I haven't figured out the cause yet

SpectMorph has two optimizations for partial screen updates, for instance for drawing a continually moving play position pointer over a sample:
* it only redraws the part of the cairo surface that was modified
* it uses GL to copy the part of the cairo surface to the screen that was modified

So it has its own version of cairo gl surface to be able to do this.

For expose events, this means that the coordinates are ignored and SpectMorph only redraws the part that has changed. Now the problem is that with your new version of the code, I get two expose events for the whole screen at the start when the window is initially shown. For the first expose event, SpectMorph renders everything. For the second expose event, it simply does nothing, because nothing was changed. The old pugl did only deliver one expose event.

So this may mean that my code is wrong, and I never should have done any partial updates in the first place. And indeed if I disable the second optimization in SpectMorph and always redraw everything for each expose event using GL then the redrawing glitches are gone.

However for a typical sample view window with a moving play position pointer the CPU usage is:
* 9.5% with both SpectMorph drawing optimizations enabled
* 18.5% with the second GL optimization disabled

So.... maybe requesting partial redraws should be pushed into pugl to allow doing this safely.

**Comment by drobilla** [comment]
That seems wrong, surely you need to draw what is expose in general?  That's what an expose is.  I could maybe make the "double expose" go away, but the whole reason that's there is to make initial display work.  There's a bunch of weird cases where it wouldn't without, but I can try to make it work...

That said, I suspect code that works like that is always going to have problems with things not being drawn when they need to be.  Pugl should deal with partial updates, and it does for ones that come from the system (this is a distinguishing thing about it, to me: it's not like some game-centric GL libs that just assume you want to draw at full blast constantly).  I guess you mean you want to post a partial update?

That would be nice.  Maybe just a general puglPostEvent() thing that can take an expose or whatever, since we already have a struct to describe exactly this and it will probably be useful to be able to post other things at some point?

**Comment by swesterfeld** [comment]
> I guess you mean you want to post a partial update?
> 
> That would be nice. Maybe just a general puglPostEvent() thing that can take an expose or whatever, since we already have a struct to describe exactly this and it will probably be useful to be able to post other things at some point?

I'm not sure if puglPostEvent() is not overdesign. A function call would do. Anyway: using expose events with an update region was my idea. So I added
```
void puglExpose(PuglView* view, int x, int y, int width, int height)
```
and used it for redrawing the area that needed updates when the play position pointer was changed. I adapted the event handling function to draw only the pixels within the expose event region.

However, the problem with my idea was that these expose events are asynchronous. So by the time the expose event was delivered, the play position pointer had already moved. So drawing only the pixels within the exposed region would usually not show the play position pointer at all, because it wasn't there any more.

So I implemented a different approach now. Instead of relying on expose events, and drawing during expose event, I now use

```
  puglProcessEvents (view);
 
  if (need_render)
    {
      puglEnterContext (view);
      render();
      need_render = false;
      puglLeaveContext (view, true);
    }
```

so after processing all events for a window, I directly update the screen using GL. So `need_render = true` is used instead of `puglPostRedisplay (view)`. This seems to work on all platforms, and I can maintain both optimizations previously used. If I really get an expose event I can simply draw everything, because this is now rare.

I retested macOS, Windows and Linux and it seems that things are now fine. Things compile (also macOS 10.9). Keys work. Redrawing glitches are no longer there. As for Windows, I believe `puglPostRedisplay` might still be buggy. It seems somehow related to your elimination of the `view->redisplay` variable, but since I no longer use it, for me things work.

So the next thing probably would be adding support for resizing my own window maybe like `puglPostResize (view, width, height)`, but thats for another PR.

**Comment by drobilla** [comment]
> I'm not sure if puglPostEvent() is not overdesign. A function call would do.

Fair enough, maybe... but I'm trying to work towards a stable API, so anything that reduces adding a bunch of functions (with maybe questionable prototypes) is appealing.  Also, what can I say, I like protocol-based design.  That said it would be a bit more annoying to call.

I guess you can do that, but drawing outside of the event loop really isn't how pugl is supposed to work (and it avoids some other optimizations), I'm not sure if it will cause problems.

It seems to me that you have two completely separate types of redraw: your internal surface (Cairo), and drawing to the window (GL).  They don't really have much to do with each other, if I understand your architecture.  So, when you get an expose event, you _must_ draw the requested part of the window (this is just a requirement of the window system).  Which means you must draw your Cairo surface to the window with GL.  You can't skip pieces here, that will definitely break things.

Then there's the question of how much updating of the surface needs to happen before drawing it.  If nothing has changed in Cairo land, then you just draw the texture and don't do anything with Cairo at all.  If something has changed in the requested region, you need to update the Cairo surface first.  This is where the playhead invalidation and such comes in.

> I believe puglPostRedisplay might still be buggy. It seems somehow related to your elimination of the view->redisplay variable, but since I no longer use it, for me things work.

It's possible, I've still mostly been testing top level windows (though I have a branch that adds embedding to pugl_test to ensure embedding cases work and stay that way), but I suspect the problems are related to the above, it seems to work fine for things that draw the requested area.

**Comment by swesterfeld** [comment]
> drawing outside of the event loop really isn't how pugl is supposed to work

Yes, I agree, and I wasn't really too happy with that approach, which is more of a workaround than a direct fix of the problem. So I've spent more time on debugging and understanding the issue, and once I really figured out what was the problem, it became surprisingly easy to fix:

The cause of all evil re-drawing bugs I observed under Linux was that before processing a configure event, I entered the GL context manually, and after processing the configure event, I left the GL context. But pugl already does this for me. So with my code, if a configure and an expose event where delivered by pugl, I left the GL context after the configure event, the expose event ran without GL context, the texture wasn't updated during expose, and redrawing started with a crappy initial state.

So long story short, it could be fixed in SpectMorph like this: 

https://github.com/swesterfeld/spectmorph/commit/502a0d27a045ab533d4e9a11f92799f99b7dee1f

and incremental redrawing under Linux is now correct. Without any new API required.

> I believe puglPostRedisplay might still be buggy. 

Ok, so I've investigated my "Windows" redrawing glitches. Some problems arose when many widgets were created/deleted. For instance in SpectMorph this happens if you load a preset. Then basically the whole UI of the old preset is deleted (think 50 widgets), and all the new widgets get created. In general, if a widget gets deleted, the UI needs to be redrawn, so I call puglPostRedisplay in the wigdet destructor. Likewise if a widget gets created, I call puglPostRedisplay in the constructor.

I could do this with the old version of pugl, because puglPostRedisplay had two properties
* it was extremely cheap (it used to set a boolean to true, and that is all it did)
* the actual redrawing was only done at the end of the event processing - so the user only saw the UI at after all creating/deleting widgets was done

The new implementation in pugl_win.c looks like this:
```
void
puglPostRedisplay(PuglView* view)
{
        RedrawWindow(view->impl->hwnd, NULL, NULL,
                     RDW_INVALIDATE|RDW_ALLCHILDREN|RDW_INTERNALPAINT);
        UpdateWindow(view->impl->hwnd);
}
```

This appears to do actual drawing, which means that all temporary state will be shown. At least I can observe that the screen is updated again an again as widgets get destroyed and created. So
* widget 1 gets deleted, calls `puglPostRedisplay`, and the screen is updated
* widget 2 gets deleted, calls `puglPostRedisplay`, and the screen is updated
* ...

So the user can watch a lot of updating going on for seconds, until the final state is reached. I was able to fix this relatively easy by replacing this with

```
void
puglPostRedisplay(PuglView* view)
{
         InvalidateRect(view->impl->hwnd, NULL, FALSE);
}
```

which doesn't do any drawing. So then the events get merged.

Since now on at least Linux and Windows `puglPostRedisplay` is no longer so cheap, it might make sense in the toolkit to collapse all redisplay operations that belong to one pugl event that is being processed into one single call. However pugl knows better if all events have been processed, so maybe the right place for collapsing `puglPostRedisplay` calls may be pugl, like it always used to be.

drobilla:
I made it so aggressive to fix problems with the window not updating initially, and in some other cases with child views.  I agree it's too much the way it is right now, but it's not so simple :/

Coalescing redisplays would be nice, but the problem with just setting a bool is that it breaks non-continuously drawing apps since the wait function will never wake up.  So _something_ needs to be posted.  _If_ it's called in an event handler, then it can work that way and they can be processed at the end, but only if.  So I think we would either need to set a bool in the view to know that, or to have different functions for this.

drobilla:
I will try to merge my more advanced pugl_test that can demonstrate this stuff soon.

PR #12, opened 2019-07-29
Set UTF8 window title using NetWM hints

jpcima:
Hi, this permits to set any UTF character in window titles on X11 platform.
It's from recent edits of DPF's pugl fork.

PS. new pugl looks great, thanks for the effort

### DIFF
--- a/pugl/detail/x11.c
+++ b/pugl/detail/x11.c
@@ -132,6 +132,9 @@ puglCreateWindow(PuglView* view, const char* title)
 
 	if (title) {
 		XStoreName(display, win, title);
+		Atom netWmName = XInternAtom(display, "_NET_WM_NAME", False);
+		Atom utf8String = XInternAtom(display, "UTF8_STRING", False);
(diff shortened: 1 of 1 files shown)


drobilla:
Cool, thanks.  As it happens I was just reading about the Windows API for this (UTF16 nightmares, whee!) so I'll add that as well and merge them together.

drobilla:
Hm, on which sort of environment can this be seen?  I changed the title in the test program first (to "Pugl Prüfung"), and... it just worked already in the XFCE in the XFCE window buttons applet.

jpcima:
MATE desktop and fr_FR.utf8 locale.
This is the "Pugl tést" window, with use of `XStoreName` only.
Before

drobilla:
It turns out that I'm just an idiot and was testing with the fix already applied :sweat_smile:

I moved the atoms since a proper place for those has been added since and fixed a const cast warning, otherwise merged more or less the same thing: https://github.com/drobilla/pugl/commit/bb0834637c7620870d784180436c90492c89db9b

I also implemented it for Windows.  This stuff should probably move to a `puglSetWindowTitle()` (this parameter was a bit silly), but this'll do for now.

Thanks.

PR #13, opened 2019-07-29
PuglEventText time in seconds

ssj71:
I'm guessing this was just a doc typo.


drobilla:
Whoops.  It got cherry-picked from an old branch with the wrong type, which @swesterfeld pointed out, but I forgot to update the comment when I fixed the type.

Merged as 6f39b67, thanks.

Issue #14, opened 2019-08-06
Blank window (Mac)


7890:
An observation on Mac OSX:
Running `pugl_test` shows a blank (white) window. The first time it's resized (if allowed), it becomes black. All keyboard and mouse entry is processed correctly BUT the scene is never drawn. Adding a printf to `onDisplay()` shows that this function is never called. The FPS stats reflect this: "0 frames in 5 seconds = 0.000 FPS". Can anybody confirm this behavior? I encounter it on macOS High Sierra Version 10.13.6. Thanks for any hints.

7890:
in `pugl_test.c` `onEvent()`, the `case PUGL_EXPOSE` never happens.

7890:
Again, it's unclear if this is the right place and correct approach to receive PUGL_EXPOSE events:
https://github.com/drobilla/pugl/blob/908e249c8030698e4313df9aeb81cfb7763e5150/pugl/detail/mac.m#L851
However adding these lines:
```
        //prevent expose when hidden
        if( ![view->impl->window isVisible] ) {return;}

        [view->impl->wrapperView setNeedsDisplay: YES];

        //create and dispatch expose event
        const NSRect rect = [view->impl->wrapperView bounds];
        const PuglEventExpose ev =  {
                PUGL_EXPOSE,
                0,
                rect.origin.x,
                rect.origin.y,
                rect.size.width,
                rect.size.height,
                0
        };

        puglDispatchEvent(view, (const PuglEvent*)&ev);
```
fixed the initial problem.

drobilla:
You can't dispatch expose events directly there.

7890:
OK, but `- (void) drawRect:(NSRect)rect` in window delegate where `PUGL_EXPOSE` is dispatched seems never called (or i missed it). I've tried with `NSLog(@"...")` to check this. Does it work at your end as currently is?

drobilla:
Your description of the issue isn't very clear.  The initial display, if continuous drawing is off, does not happen until any event occurs, yes, and then everything (not just black) draws correctly.

I am not yet sure why.  You can't fix initial display by aggressively spamming hard redraws in `puglPostRedisplay` because, aside from many other issues, it bogs down every other case.

7890:
Interesting .. 
[...]

PR #17, opened 2019-09-01
Mac: dispatch expose events from GL view drawRect

mfisher31:
Fixes broken expose events by calling dispatch from a suitable location - for #14

I noticed dispatching of configure events come from the `reshape` and `update` callbacks of the GL view.  By dispatching expose from `drawRect` subview works nicely, and also consistent with how configure events work. 

* Adds a dispatchExpose method to PuglWrapperView

### DIFF
--- a/pugl/detail/mac_gl.m
+++ b/pugl/detail/mac_gl.m
@@ -94,7 +94,7 @@ typedef NSUInteger NSWindowStyleMask;
 - (void) drawRect:(NSRect)rect
 {
 	PuglWrapperView* wrapper = (PuglWrapperView*)[self superview];
-	[wrapper dispatchExpose:[self bounds]];
+	[wrapper dispatchExpose:rect];
 }
 
 @end
--- a/pugl/detail/mac_gl.m
+++ b/pugl/detail/mac_gl.m
@@ -91,6 +91,12 @@ typedef NSUInteger NSWindowStyleMask;
 	[wrapper dispatchConfigure:[self bounds]];
 }
 
+- (void) drawRect:(NSRect)rect
+{
+	PuglWrapperView* wrapper = (PuglWrapperView*)[self superview];
+	[wrapper dispatchExpose:[self bounds]];
+}
+
 @end
 
 static int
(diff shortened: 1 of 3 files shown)


pugl/detail/mac_gl.m
```diff
@@ -91,6 +91,12 @@ - (void) update
 	[wrapper dispatchConfigure:[self bounds]];
 }
 
+- (void) drawRect:(NSRect)rect
+{
+	PuglWrapperView* wrapper = (PuglWrapperView*)[self superview];
+	[wrapper dispatchExpose:[self bounds]];
```
drobilla: The unused parameter warning here got me thinking: why bounds?  Did you choose this on purpose for some reason, rather than use the given rect?

Both seem to work fine here for me, though if there's a difference it would presumably be when subviews and such are involved...

mfisher31: > why bounds? Did you choose this on purpose for some reason, rather than use the given rect?

An oversight... should be passing `rect` if it works.  Will test on my end in just a minute.

mfisher31: Ok, tried the given `rect` works as expected. Change pushed to this P.R.

drobilla:
It seemed like there was some redundancy here, and the one `drawRect` was never called, so I dug a bit and discovered that `setNeedsDisplay` is only expected to trigger a `drawRect` for the specific view it's sent to (no parent/child stuff).

So, I propose https://github.com/drobilla/pugl/commit/f840504add3bf4a50fc4ab55db33b161d5fcb7d2 (which also includes some of the above mentioned stuff and minimizes the diff a bit)

Does that work for you as well?

mfisher31:
Yep, this is also working as expected.

drobilla:
Hm, weird.  With it this way, I _do_ get the white screen behaviour (with Cairo as well), so somehow on my machine `setNeedsDisplay` was working "recursively".

Anyway, merged as 0706f4a121f6f9c5b96a19c4272a8636fc230cbc, thanks.

mfisher31:
Odd indeed... I'll be working with pugl on a mac quite a bit over the next few weeks.  Assuming cairo-quartz isn't a nightmare to set up, I'll check it as well.

drobilla:
IIRC I got it from homebrew, it should be relatively painless

PR #19, opened 2019-09-08
x11_cairo: create double buffer with alpha content 

osch:
Create X11 cairo double buffer with CAIRO_CONTENT_COLOR_ALPHA to prevent redraw flickering while resizing windows.

Redraw flickering could occur because in `x11.c` in function `puglDispatchEvents` the redraw events are merged and delayed until  `expose.count == 0`. In `x11_cairo.c` in function `puglX11CairoResize` the backing buffer is recreated.

Now imagine the following steps: configure event for resizing comes within a series of expose events but the last expose event is not received (i.e. expouse.count is still > 0). Then we have without alpha color a recreated backing buffer filled with black color. If now the application uses `puglEnterContext`, draws some content part and then `puglLeaveContext` the drawn content part is in the backing buffer and this backing buffer is drawn on the screen. But if it is only partly drawn by the application the undrawn part is black and causes flickering. If we have alpha color in the backing buffer the undrawn part in the buffer is transparent and therefore does not change the screen content.

The above description is not an imaginary scenario, I had the described effect in a real example application.

BTW: what do you think about making the merge of exposure events configurable? The expose count is already available at the pugl api (but currently the application will not get a value != 0 because of pugl event merging). A sophisticated application could handle merging of redraw events smarter than the pugl api could. If you want, I could try preparing a pull request for optionally disabling exposure event merging.

### DIFF
--- a/pugl/detail/x11_cairo.c
+++ b/pugl/detail/x11_cairo.c
@@ -62,7 +62,7 @@ puglX11CairoCreate(PuglView* view)
 	surface.back = cairo_xlib_surface_create(
 		impl->display, impl->win, impl->vi->visual, width, height);
 	surface.front = cairo_surface_create_similar(
-		surface.back, CAIRO_CONTENT_COLOR, width, height);
+		surface.back, CAIRO_CONTENT_COLOR_ALPHA, width, height);
 	surface.backCr  = cairo_create(surface.back);
 	surface.frontCr = cairo_create(surface.front);
 
@@ -140,7 +140,7 @@ puglX11CairoResize(PuglView* view, int width, int height)
 	cairo_destroy(surface->frontCr);
 	cairo_surface_destroy(surface->front);
 	if (!(surface->front = cairo_surface_create_similar(
-		      surface->back, CAIRO_CONTENT_COLOR, width, height))) {
+		      surface->back, CAIRO_CONTENT_COLOR_ALPHA, width, height))) {
 		return PUGL_CREATE_CONTEXT_FAILED;
 	}
 


drobilla:
> If now the application uses puglEnterContext, draws some content part and then puglLeaveContext the drawn content part is in the backing buffer and this backing buffer is drawn on the screen.

Don't do this, it will cause no end of problems.

osch:
> Don't do this, it will cause no end of problems.

Why not? It works like a charm.  Anyway the application has no chance to detect this situation. Or do you suggest to never use `puglEnterContext` and `puglLeaveContext`? If the application uses these functions there is no possibility for the application to detect that a resize is in progress and that at this point there is invalid content in the backing buffer.

drobilla:
Apparently it does not work like a charm ;)

More or less, yes, applications should not use puglEnterContext, especially for drawing.  It only really exists for setting up GL, and Cairo applications should not use it at all.  Pugl is completely designed around the event loop, and the principle that drawing (among other things) happens when requested, like most toolkits.  You are getting flickering because you are circumventing the event loop and drawing somewhere/sometime you shouldn't be.

Using an imperative style and making assumptions about what the backend is doing won't work nicely, and even it it seemingly does not, it's not guaranteed to in the future.  I will improve the documentation to clarify this.

In other words, you are _required_ to draw the complete requested region in an expose.  There is no way around that, so there's no point in drawing some other time since you will need to do it again anyway.

If an application _really_ wants to draw elsewhere for some reason, it can use its own buffer and blit it during expose, but that just makes life difficult.

osch:
> Apparently it does not work like a charm ;)

Well it works like a charm if there is the alpha channel set in the backing buffer as in my pull request.

What speaks against setting the alpha channel in the backing buffer? It makes life a lot easier and is a simple and robust solution.

>  principle that drawing (among other things) happens when requested, like most toolkits

OK, but than the application needs to invoke `puglPostRedisplay` if some part of the window needs updating. But then, a exposure event for the whole window will be generated. What do you think of implementig a second function to trigger exposure event only of a certain rectangle within the window? If you want I could try to prepare a pull request for this. This features would be especially useful in conjunction with the making the exposure merging configurable as I suggested in my first post.

drobilla:
It's Complicated™

There's no end of reasons (from pugl or the system itself) why a particular region might need to be redrawn, so I think that drawing during expose is a hard requirement.  Given that, drawing at other times doesn't make much sense.  It's not very robust though, that's the problem.  It requires making all kinds of assumptions about what both the system and pugl are doing internally, and those are not always true.  The goal of the implementation is to send events in a coherent and hopefully optimal order so that things work nicely and are simple from the application's point of view.  I don't think it's really possible to have a solid Pugl implementation that works (mostly) everywhere without having applications work as intended.  In this case, "the application could have drawn to the backing buffer at any point" is not something the backend needs to be concerned with, so it doesn't need to have additional overhead to account for it.

In general, it's just been my experience that applications being weird and not working as intended always causes some problem somewhere.

> What do you think of implementig a second function to trigger exposure event only of a certain rectangle within the window? If you want I could try to prepare a pull request for this.

:+1: I think this is the right way forward.  I was thinking about doing it earlier as well, but got stuck on whether I wanted to add a bespoke function for it, or just a generic `puglPostEvent()` thing that can interpret an expose or whatever other events as they become useful in the future.

I sort of like the latter, as a general fan of minimal API surfaces and extensibility (you could even remove the resize and focus APIs in favour of that one function, which is nice), but it would be a bit more annoying to use...

osch:
OK I agree, this makes sense. I'm therefore closing this pull request now and if you want, I can try to prepare a pull request with a new function `puglPostRedisplayRect` to request redisplaying only part of the window. I could implement and test it for X11 (perhaps also Windows).

What do you think of making the "merging of exposure events by Pugl" configurable? Perhaps it's better if I create a pull request or new issue for this question.

drobilla:
Seems like it would complicate the implementation immensely, particularly when you consider configure events being in the pipe as well, and is a bit weird since the window system will do that sometimes anyway.

If it can be done cleanly, sure.

osch:
I now changed my example program and am now doing only drawing operation when I get the PUGL_EXPOSE event. However there is still some flickering when resizing the window. It seems that the expose event that belongs to the configure notify event is send after the configure notify.

xev output:
```
Expose event, serial 19, synthetic NO, window 0x5c00006,
    (0,0), width 394, height 95, count 0

ConfigureNotify event, serial 19, synthetic NO, window 0x5c00006,
    event 0x5c00006, window 0x5c00006, (4,24), width 476, height 165,
    border_width 0, above 0x0, override NO

Expose event, serial 19, synthetic NO, window 0x5c00006,
    (0,0), width 476, height 165, count 0
```
Here the expose event for width 476 comes after the configure notify for width 476. Now flickering can occur if Pugl received the first expose event for width 394 and the  ConfigureNotify for width 476 and did not yet receive the next expose event for width 476. Then short flickering can occur between ConfigureNotify and the next expose event for the same reasons I described in my first post of this thread, i.e. black areas are painted because the backing surface was recreated in the ConfigureNotify. Again everything works if I re-add the alpha channel to the backing surfe. So I'm now re-opening this pull request.

drobilla:
Of course it does, it wouldn't make sense to expose then configure for the size that's already exposed.

I do not think adding an alpha channel is an appropriate "fix".  If the double buffering is broken, then it's broken, but I don't see how this could be happening.  Can you share this program?

osch:
I'm sorry, my last post was a litte bit confusing. I'm now trying to describe it more precisely:

Take a window with white background. Then resize the window quickly using the mouse and make the window small and big rapidly. A series of Expose and ConfigureEvent will be fired as seen in the xev log of my last post. Then lets assume that after the ConfigureNotify for a short moment no more events are in the queue and Pugl therefore wants to leave `puglDispatchEvents`. Then at the end of this function there is a `pendingExpose` (with width 394, height 95) and a `pendingConfigure` (width 476, height 165). Then Pugl recreates the backing surface (it will have initially all pixels set to black) with width 476, height 165 according to the last configure event. Then the pending expose event will be processed. However this expose event will only draw the area with width 394, height 95 on the backing surface if the event's exposure coordinates are used by the event callback. The rest of the backing area contains black pixels. Then these black pixels are drawn on the screen. With this black content is shortly shown while resizing whereas the eye does expect white color. This could be perceived as some kind of flickering.

One could fix this by expanding the expose event coordinate to the full view if pending configure and expose event is processed at the same time. But then the next expose event will come with width 476, height 165 and the screen is drawn again.

By setting the alpha channel on the backing surface everything works as it should: only the part of backing surface where the application has drawn something is transfered to the screen, the fully transpart parts are omitted. Therefore only pixels are going to the screen in the area where a expose event was received.

The need for the alpha channel becomes more important if there is also a method `puglPostRedisplayRect` for which I am preparing a pull request right now: Assume that the user resizes the window as described above and the application does at the same time invoke `puglPostRedisplayRect` (e.g. for animating a part of the window). Then there are ConfigureNotify events and Expose events resulting from the resize operation and Expose events send by the application intermixed. With alpha channel on the backing surface everything works like it is designed: the application draws the part for which Pugl accumulated expose events and because of the alpha channel only the things that are drawn on the backing surface by the application are transfered to the screen. For this it doesn't matter when the backing surface was recreated.

osch:
I'm closing this pull request again, since these changes are contained in #20.

osch:
> Rephrased, this means "parts of the exposed window are not drawn".

but these parts are not "exposed" in my example. They will be later on if the next expose events are received. On the other hand you are right: if the application does not draw the corresponding area of an expose event, garbage will be left on the screen. But this can also happen without alpha channel if the backing buffer is freshly created (then black pixels are the "garbage") or if the expose event results from "puglPostRedisplay" (then the "garbage" is old content that the applicaion has drawn before).

> That is why alpha blending a back buffer doesn't make sense: alpha blending it onto what? 
Garbage, potentially.

Alpha blending the backing buffer could IMHO also make sense for other scenarios: imagine a X11 view embedded into a larger native X11 window (both things are called "Window" in X11 terminolgy: a top level window and an embedded sub window). The parent window could draw a background behind the embedded view. If the embedded view draws transparent content, the background of the window will be visible as background in the embedded view. This could be a desired effect for having a uniform background behind subviews.


> If this happens, that is the problem.

but this problem can be solved as I wrote. IMHO this is the simplest solution: just respecting the x11 event queue and only transfer things to the screen that are exposed and drawn at this point in the event queue. If you are concerned that the application does not properly draw content in the expose callback and garbage could appear: Pugl could optionally clear the areas in the backing buffer that are exposed with a background color. I would make this optionally because IMHO the correct way is that the application does correct drawing.

As I'm writing this I'm getting the idea that the double buffer could be seen from two point of views:

1. using the backing buffer to minimize drawing: Then Pugl must distinguish if the expose event comes because the application wants to draw changed content or because some area of the screen becomes invalid and old content just must be repainted: however these days the X11 server does the latter most of the times: X11 server have backing stores enabled such that there are no expose events send if for example another window is moved away and parts of the window behind become visible (again).

2. using backing buffer to minimize screen flickering: The application does some drawing into the backing buffer that will be transfered to the screen at once. For this scenario it is not needed that the backing buffer must contain the whole content that the application did ever draw. It would just be enough to hold the content that must be exposed in this moment.

osch:
As a proof of concept see also #22: this shows that it is not necessary to 100% track the window content in a backing surface.

drobilla:
I can't reproduce this issue.  If I make the background white in pugl_cairo_test, for example, resizing is perfectly smooth and the whole background is always white.

Can you please try this, and/or provide a program that demonstrates the problem, and be more specific about your setup (WM in particular)?

drobilla:
> IMHO this is the simplest solution: just respecting the x11 event queue and only transfer things to the screen that are exposed and drawn at this point in the event queue.

Conceptually, sure, that's the simplest thing, and how it was originally written, but it makes resizing really terrible.

osch:
>  provide a program that demonstrates the problem

It was not so easy to reproduce this with `pugl_cairo_test` but finally I managed it: see https://github.com/osch/pugl/commit/318f903aad389f3d95f2c0faf2a89a174fd6e1d0

Start `pugl_cairo_test1` without option `-c`.  On faster machine the loop for `i` must be increased.

However while testing this I found an error in `puglPollEvents`, see #23.  This error produced the gap between the last ConfigureEvent and the following ExposeEvent. In theory (e.g. remote x11) there could be a gap, but with this fix it is nearly impossible to reproduce this gap.

drobilla:
Ah, I see.  It wasn't using the expose size anyway.  Thanks, I'll look into it.

PR #20, opened 2019-09-09
new function puglPostRedisplayRect

osch:
As discussed in #19 here comes a pull request for requesting redisplaying a rectangular part of a PuglView using the new function:

```
PUGL_API PuglStatus
puglPostRedisplayRect(PuglView* view, PuglRect rect);
```

I tested this functionality thoroughly under X11 (local and remote display), Win and Mac. For Mac some minor fixes were necessary.

Edit: I tested only Cairo not OpenGL.

### DIFF
--- a/pugl/detail/mac.m
+++ b/pugl/detail/mac.m
@@ -946,6 +946,14 @@ puglPostRedisplay(PuglView* view)
 	return PUGL_SUCCESS;
 }
 
+PuglStatus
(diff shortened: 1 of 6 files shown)


drobilla:
I decided the redisplay is inherently problematic, so took a slightly different approach and just removed it entirely.  Said function is added in be838db, thanks.

Issue #18, opened 2019-09-07
PuglWorld: access for Display* 


osch:
Hi, I'm very interested in the latest development for multi view pugl.

For the `PuglView` the native window handle can be accessed with  `puglGetNativeWindow`. 

What do you think about implementing a similar access to the underlying native handle for the `PuglWorld`? For X11 this would be the `Display*`. By looking at the code for the other plattforms it seems to me that there is nothing similiar for other platforms (NSApplication on mac is singleton). See also my implementation of `puglGetNativeDisplay` for X11: https://github.com/osch/pugl/commit/59e03dd1bf22ff0deca561422d371dcb2a31530e

Such access could be useful for a lot of things but I need it especially for creating pixmap cairo surfaces on the X11 server independently from the views. This could be solved externally if the `Display*` can be reached. But it could also be a good idea to integrate this feature into `PuglWorld`: a new function to create a cairo pixmap surface independently from the PuglViews that will match the current screen device and will be on the display server for x11. What do you think of this approach?

drobilla:
> By looking at the code for the other plattforms it seems to me that there is nothing similiar for other platforms

It's worse, there is similar Cairo functionality for all the platforms, but all of the others require an actual drawing context to create a compatible surface.  This pattern itself is non-portable.

> I need it especially for creating pixmap cairo surfaces on the X11 server independently from the views

Why?

> a new function to create a cairo pixmap surface independently from the PuglViews that will match the current screen device and will be on the display server for x11. What do you think of this approach?

I don't think it's appropriate for the Pugl API to have X11 and Cairo specific things in it.  Maybe if this concept could be mapped to some reasonable portable API and implemented by backends, and it made sense for arbitrary backends, but that seems questionable, and doesn't fit the design where backends are part of views anyway.

Exposing the display isn't so bad, and it's simple, so I'm not opposed to it, but it smells a bit since the concept doesn't map to the other platforms and it seems like the thing you are doing fundamentally doesn't work on them at all, so I'd like to understand why you want to create a server-side surface but have no view to base it on.

osch:
> I don't think it's appropriate for the Pugl API to have X11 and Cairo specific things in it.

Yes in principle the normal Pugl API should be platform agnostic but in more complicated cases the application could also have some platform specific parts and for these cases it could be useful to get acccess to some specific details in pugl. One idea could be to  provide for theses cases some public platform specific functions, e.g. just having a non static function `puglX11GetDisplay` in `x11.c`.

[...]

PR #21, opened 2019-09-09
puglX11GetDisplay

osch:
Here comes a implementation for `puglX11GetDisplay` as discussed in #18.


drobilla:
Replaced with `puglGetNativeWorld()` for interface portability in d53ce36, thanks.

PR #22, opened 2019-09-10
x11_cairo: minimize x11 server side memory consumption

osch:
This minimizes x11 server side memory consumption when using cairo (you can examine it with `xrestop`). I didn't notice a performace decrease for example using `test/pugl_cairo_test.c` with options `-a -c -r`.

### DIFF
--- a/pugl/detail/x11_cairo.c
+++ b/pugl/detail/x11_cairo.c
@@ -51,6 +51,59 @@ puglX11CairoConfigure(PuglView* view)
 	return PUGL_SUCCESS;
 }
 
+static void
+puglX11CairoClose(PuglView* view)
+{
+	PuglInternals* const       impl    = view->impl;
+	PuglX11CairoSurface* const surface = (PuglX11CairoSurface*)impl->surface;
+
+	if (surface->frontCr) {
+		cairo_destroy(surface->frontCr);
+		surface->frontCr = NULL;
+	}
+	if (surface->front) {
+		cairo_surface_destroy(surface->front);
+		surface->front = NULL;
+	}
+	if (surface->backCr) {
+		cairo_destroy(surface->backCr);
+		surface->backCr = NULL;
+	}
+	if (surface->back) {
+		cairo_surface_destroy(surface->back);
+		surface->back = NULL;
+	}
+}
+
+static PuglStatus 
+puglX11CairoOpen(PuglView* view)
+{
+	PuglInternals* const       impl    = view->impl;
+	PuglX11CairoSurface* const surface = (PuglX11CairoSurface*)impl->surface;
+
+	puglX11CairoClose(view); // just to be sure
+	
+	surface->back = cairo_xlib_surface_create(
+		impl->display, impl->win, impl->vi->visual, 
+		view->frame.width, view->frame.height);
+	
+	if (surface->back) {
+		surface->backCr = cairo_create(surface->back);
+		surface->front = cairo_surface_create_similar(
+			surface->back, CAIRO_CONTENT_COLOR_ALPHA, 
+			view->frame.width, view->frame.height);
+	}
+	if (surface->front) {
+		surface->frontCr = cairo_create(surface->front);
+	}
+	if (surface->frontCr) {
+		return PUGL_SUCCESS;
+	} else {
+		puglX11CairoClose(view);
+		return PUGL_CREATE_CONTEXT_FAILED;
+	}
+}
+
 static PuglStatus
 puglX11CairoCreate(PuglView* view)
 {
@@ -59,26 +112,12 @@ puglX11CairoCreate(PuglView* view)
 	const int            height  = (int)view->frame.height;
 	PuglX11CairoSurface  surface = { 0 };
 
-	surface.back = cairo_xlib_surface_create(
-		impl->display, impl->win, impl->vi->visual, width, height);
-	surface.front = cairo_surface_create_similar(
-		surface.back, CAIRO_CONTENT_COLOR, width, height);
-	surface.backCr  = cairo_create(surface.back);
-	surface.frontCr = cairo_create(surface.front);
+	surface.back    = NULL;
+	surface.backCr  = NULL;
+	surface.front   = NULL;
+	surface.frontCr = NULL;
 
 	cairo_status_t st = CAIRO_STATUS_SUCCESS;
-	if (!surface.back || !surface.backCr ||
-	    !surface.front || !surface.frontCr ||
-	    (st = cairo_surface_status(surface.back)) ||
-	    (st = cairo_surface_status(surface.front)) ||
-	    (st = cairo_status(surface.backCr)) ||
-	    (st = cairo_status(surface.frontCr))) {
-		cairo_destroy(surface.frontCr);
-		cairo_destroy(surface.backCr);
-		cairo_surface_destroy(surface.front);
-		cairo_surface_destroy(surface.back);
-		return PUGL_CREATE_CONTEXT_FAILED;
-	}
 
 	impl->surface = calloc(1, sizeof(PuglX11CairoSurface));
 	*(PuglX11CairoSurface*)impl->surface = surface;
@@ -92,10 +131,7 @@ puglX11CairoDestroy(PuglView* view)
 	PuglInternals* const       impl    = view->impl;
 	PuglX11CairoSurface* const surface = (PuglX11CairoSurface*)impl->surface;
 
-	cairo_destroy(surface->frontCr);
-	cairo_destroy(surface->backCr);
-	cairo_surface_destroy(surface->front);
-	cairo_surface_destroy(surface->back);
+	puglX11CairoClose(view);
 	free(surface);
 	impl->surface = NULL;
 	return PUGL_SUCCESS;
@@ -104,15 +140,10 @@ puglX11CairoDestroy(PuglView* view)
 static PuglStatus
 puglX11CairoEnter(PuglView* view, bool drawing)
 {
-	PuglInternals* const       impl    = view->impl;
-	PuglX11CairoSurface* const surface = (PuglX11CairoSurface*)impl->surface;
-
-	if (drawing) {
-		cairo_save(surface->frontCr);
-	}
-
 	return PUGL_SUCCESS;
 }
+
+
 
 static PuglStatus
 puglX11CairoLeave(PuglView* view, bool drawing)
@@ -120,11 +151,11 @@ puglX11CairoLeave(PuglView* view, bool drawing)
 	PuglInternals* const       impl    = view->impl;
 	PuglX11CairoSurface* const surface = (PuglX11CairoSurface*)impl->surface;
 
-	if (drawing) {
+	if (drawing && surface->frontCr) {
 		cairo_set_source_surface(surface->backCr, surface->front, 0, 0);
 		cairo_paint(surface->backCr);
-		cairo_restore(surface->frontCr);
 	}
+	puglX11CairoClose(view);
 
 	return PUGL_SUCCESS;
 }
@@ -132,21 +163,7 @@ puglX11CairoLeave(PuglView* view, bool drawing)
 static PuglStatus
 puglX11CairoResize(PuglView* view, int width, int height)
 {
-	PuglInternals* const       impl    = view->impl;
-	PuglX11CairoSurface* const surface = (PuglX11CairoSurface*)impl->surface;
-
-	cairo_xlib_surface_set_size(surface->back, width, height);
-
-	cairo_destroy(surface->frontCr);
-	cairo_surface_destroy(surface->front);
-	if (!(surface->front = cairo_surface_create_similar(
-		      surface->back, CAIRO_CONTENT_COLOR, width, height))) {
-		return PUGL_CREATE_CONTEXT_FAILED;
-	}
-
-	surface->frontCr = cairo_create(surface->front);
-	cairo_save(surface->frontCr);
-
+	puglX11CairoClose(view);
 	return PUGL_SUCCESS;
 }
 
@@ -156,6 +173,9 @@ puglX11CairoGetContext(PuglView* view)
 	PuglInternals* const       impl    = view->impl;
 	PuglX11CairoSurface* const surface = (PuglX11CairoSurface*)impl->surface;
 
+	if (!surface->frontCr) {
+		puglX11CairoOpen(view);
+	}
 	return surface->frontCr;
 }
 


drobilla:
I did a quick test by disabling the text rendering in pugl_cairo_test (which is a huge bottleneck) and this change drops the FPS by about 150, which isn't too surprising since it adds allocating surfaces to every frame.

Caching the context like master does currently is probably mostly pointless (it's designed to be fast to create and save/restore is mostly the same thing anyway), but I don't see a reason to constantly allocate entire buffers.

osch:
I measured on two machines: master without text rendering is slightly faster on the second machine. For the others: the new rendering with minimized memory seems to be even faster sometimes. But I think this is not a clear result. So it seems to me that in average minimizing the server memory does not have a relevant performance drawback.

Machine 1, master, with text rendering, 
```
7836 frames in 5 seconds = 1567.105 FPS
7580 frames in 5 seconds = 1515.977 FPS
7670 frames in 5 seconds = 1533.966 FPS
7848 frames in 5 seconds = 1569.470 FPS
7809 frames in 5 seconds = 1561.591 FPS
7682 frames in 5 seconds = 1536.391 FPS
```

Machine 1, min-mem, with text rendering
```
5993 frames in 5 seconds = 1198.497 FPS
7875 frames in 5 seconds = 1574.821 FPS
8198 frames in 5 seconds = 1639.514 FPS
8116 frames in 5 seconds = 1623.113 FPS
8203 frames in 5 seconds = 1640.505 FPS
8109 frames in 5 seconds = 1621.767 FPS
```

Machine 1, master, without text rendering
```
10588 frames in 5 seconds = 2117.495 FPS
10579 frames in 5 seconds = 2115.742 FPS
10659 frames in 5 seconds = 2131.619 FPS
10506 frames in 5 seconds = 2101.129 FPS
```

Machine 1, min-mem, without text rendering
```
10511 frames in 5 seconds = 2102.109 FPS
10581 frames in 5 seconds = 2116.173 FPS
10658 frames in 5 seconds = 2131.382 FPS
10550 frames in 5 seconds = 2109.859 FPS
```

Machine 2, master, with text rendering
```
3425 frames in 5 seconds = 684.887 FPS
3350 frames in 5 seconds = 669.988 FPS
3696 frames in 5 seconds = 739.131 FPS
3433 frames in 5 seconds = 686.589 FPS
3533 frames in 5 seconds = 706.578 FPS
```
Machine 2, min-mem, with text rendering
```
3842 frames in 5 seconds = 768.333 FPS
3678 frames in 5 seconds = 735.430 FPS
3607 frames in 5 seconds = 721.216 FPS
3609 frames in 5 seconds = 721.658 FPS
3734 frames in 5 seconds = 746.759 FPS
```

Machine 2, master, without text rendering
```
6274 frames in 5 seconds = 1254.719 FPS
6118 frames in 5 seconds = 1223.462 FPS
6118 frames in 5 seconds = 1223.447 FPS
6237 frames in 5 seconds = 1247.338 FPS
6330 frames in 5 seconds = 1265.970 FPS
```
Machine 2, min-mem without text rendering
```
5637 frames in 5 seconds = 1127.273 FPS
5684 frames in 5 seconds = 1136.789 FPS
5848 frames in 5 seconds = 1169.582 FPS
5760 frames in 5 seconds = 1151.833 FPS
5838 frames in 5 seconds = 1167.594 FPS
```

osch:
So do think that minimizing the x11 server side memory is a good idea or not? 

If there is not a permanent animation on the whole window the server side memory usage is zero and the performance seems to be even slightly better in the worst case for permanent full window redrawing, so IMHO this PR does overall improve things.

drobilla:
I see no point at all in "minimizing" server side memory _sometimes_, when two buffers are going to exist at the same time anyway.

Which means the fastest things wins, which seems is the current master.  I find your results with text rendering being faster... odd.  Why would that be the case?  I get the opposite, master is always faster, which makes sense because it's doing much less.

osch:
> I see no point at all in "minimizing" server side memory sometimes, when two buffers are going to exist at the same time anyway.

My impression is, that x11 server and cairo are optimized to handle these things themselfs and there is no need for the client to permanently reserve buffers only for temporary usage. 

If you have multiple pugl windows in current master each one will allocate server side memory permanently. With the minimized memory patch the overall memory usage will be much lower because memory will be reused for different windows.

Look at `xrestop` with a maximized Pugl window on current master: this will permanently allocate more memory than most gtk applications with the same window size are using. If one application has more Pugl windows this memory will linearly increase whereas in most gtk applications the memory does not increase lineary for more windows. So I'm pretty shure that gtk itself does not allocate permanent buffers for each window.

One could improve Pugl by only allocating the buffer in `PuglWorld` and share it between `PuglViews`. But I don't see an advantage here because the performace gain for permanently allocating buffers seems to be minimal for typical applications, and on my machines it is even faster without permanent buffers.  What frame rates did you measure? You mentioned only a drop of 150.

>  I find your results with text rendering being faster... odd. Why would that be the case?

I assume cairo could do better optimization for this by reusing memory and/or graphic card resources that are already used by other applications. With permanently allocated buffers the x server perhaps must do more switches to use the special reserverd resources. What version of cairo and x server do you have? I have cairo 1.17.2 and x server 1.20.5.

drobilla:
Cairo 1.17.3 Xorg 1.20.5

Sorry, I would need to see some positive results myself to consider this.  I don't care what xrestop says, to be honest.  As far as I can tell it just makes everything slower (and Pugl is not intended for apps that create a ton of windows anyway).  What I see - that doing more work by allocating buffers every frame makes things slower - is intuitive and makes sense.

osch:
> What I see - that doing more work by allocating buffers every frame makes things slower - is intuitive and makes sense.

my intuition says different: it's often better to have small memory and resource usage for better performance.

drobilla:
This does not reduce resource consumption in the case of drawing...

osch:
To get a better understanding I did some more measurements:

1. one app `pugl_cairo_test` with `-a -c -r` and different sizes (512x512 and maximized to screen 1920x1080):

```
master 512x512:    1456 FPS
master maximized:   390 FPS

minmem 512x512:    1563 FPS
minmem maximized:   317 FPS
```

2. four simultanously running apps `pugl_cairo_test` (original size 512x512) with `-a -c -r`:
```
master: 273 FPS per app
minmem: 319 FPS per app
```

These numbers indicate that on this machine smaller windows seems to be rendered from the x server in some kind of (hardware?) accelerated context that is shared between apps whereas bigger windows are not running in the accelerated context and for minmem then there is some overhead for always creating a new buffer. 

So IMHO from this there is still no clear indication if minmem is general a good idea or not, but it shows that in the minmem case the x server could do (hardware?) acceleration, if the buffer is not too large, wheres in the case with fixed buffers there is no acceleration. I would guess that the above results also depend on the graphics card (on this machine it was: "Intel Xeon E3-1200 v3/4th Gen Core Processor Integrated Graphics Controller").

drobilla:
After entirely too much time, I finally figured out why the alpha kludge is required to make this work.  You're right that Gtk does it this way, but there is no alpha there, and they even use CAIRO_OPERATOR_SOURCE in addition.  Doing that in Pugl makes everything horrible though, especially in partial expose scenarios highlighted by pugl_cairo_test.

The "why" is obvious in retrospect: there's no clipping.  Even ignoring the double buffering strategy part, the current X11 Cairo backend is totally wrong for partial exposures.

I'm working on fixes for this and a rework of this change on top of it.  This will change the performance situation a bit, but it seems to work well for me.  Still slower, but it will be interesting to see what the effect is on that strange machine of yours...

drobilla:
Try https://github.com/lv2/pugl/tree/partial-draw

osch:
OK, I tested on three machines: Machines 1 & 2 are Thinkpads (ca. >= 6 years old), 3 is a normal PC (also ca. >= 6 years old).

  * Test A: current Master
  * Test B: current Master with my alternative x11 cairo backend (see https://github.com/osch/pugl/blob/test-x11-cairo/pugl/detail/x11_cairo_01.c)
  * Test C: your `partial-draw´ branch (https://github.com/lv2/pugl/tree/partial-draw)
 
Test program is `pugl_cairo_test  -a -c -r`

Results:

  * Machine 1
    * Test A: 2229.772 FPS
    * Test B: 2474.833 FPS
    * Test C: 774.261 FPS

  * Machine 2
    * Test A: 2907.769 FPS
    * Test B: 2895.188 FPS
    * Test C: 1776.110 FPS

  * Machine 3
    * Test A: 3098.664 FPS
    * Test B: 5367.722 FPS
    * Test C: 1603.596 FPS

Best regards,
Oliver

drobilla:
Sorry, I should have clarified.  The frame rates of this branch are not comparable to the others because it used to be counting exposures which were not merged (4 per frame).

osch:
> The frame rates of this branch are not comparable to the others

And how did you compare performance?

drobilla:
https://github.com/lv2/pugl/commit/2495d2deb77c8474d7bbe467a0d033909ee3263c is the relevant commit.

osch:
I'm sorry but I don't understand your answer. 2495d2d is the partial redraw branch (except that afterwards pugl_cairo_test.c got slight modifications. I tested on this branch. You wrote "It is a bit slower for a single application in most cases". So how did you compare the performance and what did you measure? Did you also measure my approch from https://github.com/osch/pugl/blob/test-x11-cairo/pugl/detail/x11_cairo_01.c ?

drobilla:
Compare which performance?  If you interested in the memory consumption / ephemeral surface thing, then that commit is the one that does that.  Measuring before and after that commit will show the performance with long-lived vs short-lived Cairo surfaces.

The expose model on old master is just wrong, there is no comparison to be made between master (or your original branch on it) and partial-draw.  Your original alpha kludge was a workaround for this deeper problem, but even with that, it resulted in a bunch of visibly sketchy behavior to try to draw more than is actually exposed.

That said, at least for me the "draw as fast as possible" thing that pugl_cairo_test does completely bogs down the X server and stalls other instances of the program when there's lots of events (like mouse movement), so I'm not sure how much benchmarking this program as it is makes sense anyway.  I don't know if, or even how, it's even possible to draw as fast as possible on X11, or at least the event loop API would probably need to be changed to be able to write a reasonable real program that works this way.

Anyway, it's simpler and what Gtk does and not significantly slower (and Cairo is slow regardless) so I don't really care much.

osch:
> Your original alpha kludge was a workaround for this deeper problem, but even with that, it resulted in a bunch of visibly sketchy behavior to try to draw more than is actually exposed.

Do you have more details on this? I tested it thoroughly and didn't notice any flaws. However I always assumed that the application is setting up a clipping region on the context it is drawing.

Unfortunately the current solution enforces a specific drawing model on the application. I'm experimenting with the possibility that the application handles the list of individual exposure rectangles itself and not only one merged exposure rectangle. Roughly speaking I see two types of drawing models:
  1. simple application that always just redraws the whole view at once: for these applications it is best just to get one merged exposure rectangle. It makes no sense for this type of application to trigger partial redraws.
  2. application that considers redraw rectangles and draws only affected widgets: this type of application would ideally collect all redraw rectangles and make one drawing pass through the widgets considering the whole list of rectangles.

You may have a look at https://github.com/osch/pugl/commit/432196659691ececf1afb66b00187ac17070830c where I implemented an option to have support for both types of drawing models (even for Windows and MacOS < 10.14) by passing the individual exposure rectangles to the application using the `count` member of the exposure event structure.

There seems to be a use case for drawing model 2. in real-world scenarios, see e.g. https://forum.juce.com/t/juce-coregraphics-render-with-multiple-paint-calls-not-working-on-new-mac-mojave/30905 or https://github.com/steinbergmedia/vstgui/commit/245ba880aa807a17aafac515efaeaa92c13f5093

drobilla:
> Do you have more details on this? I tested it thoroughly and didn't notice any flaws. However I always assumed that the application is setting up a clipping region on the context it is drawing.

It doesn't matter if you are clipping in the application if pugl is blitting the entire surface anyway.  Even worse if it's doing so with alpha blending.

You can't immediately dispatch expose events like pugl was before.  It conflicts with the deferred expose model where exposes happen at the end of the loop, and generally causes a bunch of weird problems with all these exposes coming from different places and happening at different times.  Most importantly and especially on X11, it also means that the exposes probably won't happen in this event loop at all, which is a real problem people have had that causes noticeable lag, especially in plugin UIs where things are not driven so quickly.  It means that if you request a redraw of your animating button or whatever, the actual expose might not come back until a whole event loop later.  

Not unifying isn't a different model so much (you are always required to draw whatever the library tells you to, no exceptions), it's just an optimization.  If you want to implement that, awesome, but as I say in the commit, for that pugl would have to implement a rect set instead of storing just the one, probably add optimization of it to not send overlapping exposures, etc etc.

drobilla:
Analogous change made in https://github.com/lv2/pugl/commit/2495d2deb77c8474d7bbe467a0d033909ee3263c

osch:
>It doesn't matter if you are clipping in the application if pugl is blitting the entire surface anyway. 

The difference it that if it's clipped in the application the clipped areas will remain alpha = 0.0 even if some widgets are drawing something outside the clipping region. So if there are widgets that draw outside of the drawing region than there should be some clipping. For the end result it doesn't matter if the clipping is done in the application or on the x11 target surface. If the widgets do consider the clipping area than there would be no clipping necessary. If clipping is necessary than it's better to do it on the drawing surface since otherwise things are drawn on the drawing surface that may have been optimized away by cairo's clipping and only afterwards these drawings are discarded if the drawing surface in blitted onto the x11 target surface.

> Even worse if it's doing so with alpha blending.

Why do you think it's worse? All the measurements I made on three different machines showed that this is not the case (at least in my cases), i.e. there seems to be optimization for blending larger areas with alpha = 0.0. It's an implementation detail within cairo wether clipping performance is better than areas with alpha = 0.0 and IMHO you can only decide this if you study cairo's source code or do some measurements.

> Most importantly and especially on X11, it also means that the exposes probably won't happen in this event loop at all, which is a real problem people have had that causes noticeable lag,

Yes I know  and considered this in my implementation.

> If you want to implement that, awesome

I already implemented it, see https://github.com/osch/pugl/commit/432196659691ececf1afb66b00187ac17070830c

> for that pugl would have to implement a rect set instead of storing just the one

This would be one possible solution. I chose another way that does better fit into current pugl's architecture by letting the application collect the rectangle list. The application can choose the drawing model 2 by setting `puglview->hints[PUGL_DONT_MERGE_RECTS])` and after this the exposure rectangles are not merged but are delivered to the application. With this option set the application should only draw if `count == 0` in the exposure event. As I wrote: this does also work under Windows and Mac OS < 10.14.

> Analogous change made in 2495d2d
> drobilla closed this 1 hour ago

Yes I agree the purpose of this issue has been reached: x11 server side memory is minimized.

If you are interested in further collaboration on supporting drawing model 2 I could open new issue or PR.

drobilla:
I am not interested in discussing this alpha thing anymore.  It was an ugly kludge around drawing being broken, drawing a larger region with alpha blending is obviously worse than drawing a smaller one without it, clipping in the application is a separate thing, and drawing is no longer broken so it isn't needed anyway.

> Yes I know and considered this in my implementation.

Apparently not https://github.com/osch/pugl/commit/432196659691ececf1afb66b00187ac17070830c#diff-ee74428ad181f3c9964f8ea87536619bR1051

> This would be one possible solution.

That is the only possible solution.

osch:
>>Yes I know and considered this in my implementation.

> Apparently not osch@4321966#diff-ee74428ad181f3c9964f8ea87536619bR1051

you are refering to:

https://github.com/osch/pugl/blob/432196659691ececf1afb66b00187ac17070830c/pugl/detail/x11.c#L1046-L1051
```C
		XExposeEvent ev = {Expose, 0, True,
		                   view->impl->display, view->impl->win,
		                   x, y,
		                   w, h,
		                   0};
		XSendEvent(view->impl->display, view->impl->win, False, 0, (XEvent*)&ev);
```
but this is the `else` case of the following `if`: please have a closer look at:

https://github.com/osch/pugl/blob/432196659691ececf1afb66b00187ac17070830c/pugl/detail/x11.c#L1030-L1039

```C
puglPostRedisplayRect(PuglView* view, PuglRect rect)
{
	if (view->world->impl->dispatchingEvents) {
		// Currently dispatching events, add/expand expose for the loop end
		const PuglEventExpose event = {
			PUGL_EXPOSE, 0, rect.x, rect.y, rect.width, rect.height, 0
		};


		addPendingExpose(view, (const PuglEvent*)&event);
	} else if (view->visible) {
```
so I'dont think that it's apparent that I didn't consider it. Or are you seeing something that I'm missing? Than I would be happy to hear more details.

drobilla:
https://github.com/osch/pugl/blob/432196659691ececf1afb66b00187ac17070830c/pugl/detail/x11.c#L840

osch:
You are referring to the invocation of `puglDispatchEvent` in https://github.com/osch/pugl/blob/432196659691ececf1afb66b00187ac17070830c/pugl/detail/x11.c#L834-L847

```C
addPendingExpose(PuglView* view, const PuglEvent* expose)
{
        if (view->hints[PUGL_DONT_MERGE_RECTS]) {
            if (view->impl->pendingExpose.type) {
                puglEnterContext(view, true);
                view->impl->pendingExpose.expose.count += 1 + expose->expose.count;
                puglDispatchEvent(view, &view->impl->pendingExpose); //<<<<<<
                puglLeaveContext(view, true);
            }
            view->impl->pendingExpose = *expose;
        } else {
            mergeExposeEvents(&view->impl->pendingExpose, expose);
        }
}
```

This is the part where my given description comes in: 

>The application can choose the drawing model 2 by setting puglview->hints[PUGL_DONT_MERGE_RECTS]) and after this the exposure rectangles are not merged but are delivered to the application. With this option set the application should only draw if count == 0 in the exposure event.

I cannot see what is wrong here except that the invocation of `puglEnterContext(view, true);` could be questioned, since the application should only collect the rectangle here and should not draw (because `count>0`). So perhaps it's better to omit `puglEnter/LeaveContext` here or to pass `false` instead of `true`.

I'm not saying that this is the best thinkable or only solution here but at least it is a working starting point. I'm open also for others solutions/suggestions.

drobilla:
Making the application collect rectangles to draw... at.. some point.. later... (?) is confusing and error-prone at best.  That puts the burden on applications to deal with all of this correctly (which they won't), and I don't even see how it would be possible to do so.  If you don't draw when you get expose events then when do you draw?  Pugl now needs to send some other expose event later?  With what region?  We add some weird `PuglReallyExposePendingExposuresNow` event?

That's _way_ too complicated, especially if the semantics are different between platforms.  The entire point of a library like Pugl is to abstract away this kind of crap.

(What's the point anyway?  You don't need Pugl to send changed rectangles to yourself, if you _really_ want to draw to an image and update only the parts you know have changed and expose that instead, you can).

This is the Pugl expose API:

* The application is required to draw the entire expose region when handling that event.
* Any posted redisplays in the event loop produce exposures which are dispatched at the end of the same `puglDispatchEvents` invocation.

Simple, universal/well-understood, and it's easy to write both applications and backends correctly based on this model.  It won't change.

I'm still not sure if this is _guaranteed_ on Windows or MacOS using the window system facilities.  If not, those backends will have to handle it internally like the X11 one does.  I'd look into this before investing more time into using InvalidateRect and friends.

> I'm open also for others solutions/suggestions.

Again, Pugl needs to maintain a rect list to do this, at least on X11.  If there's platform support, like you used on Windows there, and the semantics are right, then that's fine too, but no breaking the drawing model.

osch:
>  That puts the burden on applications to deal with all of this correctly (which they won't), 

Drawing model 2 is only for application that want to handle this correctly, i.e. for applications that are dealing with a lot of smaller widgets and decide from the list of exposure rectangles which widget is to be drawn. To make this efficient it is best to collect the list of rectangles and let the applications handle them at once. For Windows and Mac the list of rectangles can be obtained from the operating system when the redraw event occurs.

> That's way too complicated, especially if the semantics are different between platforms.

In my implementation the semantics is the _same_ for all platforms: if the application sets `PUGL_DONT_MERGE_RECTS` then the application must only draw on exposure events with `count == 0` and for `count > 0` only should add the rectangle to a list. IMHO that's a simple solution for drawing model 2. IF the `PUGL_DONT_MERGE_RECTS` is not set, then every exposure event will be with `count == 0` and the application must draw on every exposure event.

> if you really want to draw to an image

I don't want to draw to an image.

> Pugl now needs to send some other expose event later?

No, it's an exposure event with `count == 0`.

drobilla:
> To make this efficient it is best to collect the list of rectangles and let the applications handle them at once.

At once _when_?

drobilla:
Judging by `pugl_redisplay_test` in https://github.com/lv2/pugl/tree/test, MacOS can't be relied on to dispatch these immediately anyway, so pugl has to handle it there too.

I wouldn't be surprised if the same is true on Window as well but it's just less likely, but I'm not sure.

osch:
> Which is sent when?

For X11 it is (as it was before) at the end of the function `puglDispatchEvents`, see  https://github.com/osch/pugl/blob/432196659691ececf1afb66b00187ac17070830c/pugl/detail/x11.c#L992-L994

For Win it is the remaining expose event with the last recangle after all other rectangles have been dispatched, see https://github.com/osch/pugl/blob/432196659691ececf1afb66b00187ac17070830c/pugl/detail/win.c#L855-L865

For Mac it is the last expose event that is dispatched, see https://github.com/osch/pugl/blob/432196659691ececf1afb66b00187ac17070830c/pugl/detail/mac_cairo.m#L66-L68

> MacOS can't be relied on to dispatch these immediately anyway

Immediate dispatch of exposure events is for Win and Mac not as necessary as it is for X11 because on Win and Mac this is all happening in the central event queue that is also driven outside the PuGL event loop. On X11 there is no central event queue so if on X11 the plugin does not start its own GUI thread but only handles events in an idle callback than there might be unresponsive behaviour if events are stuck in the event queue and the idle callback is only invoked again after a certain time passed.

drobilla:
Then how do I know whether to draw the entire region that expose specifies, or only the ones that came before?

What happens when the window system sends exposes too?  What if they overlap?

What is the advantage of all this complexity from the point of view of the application developer?

7890:
Slightly related, how could application developer inject a custom event from application logic to be handled like any other PUGL_ event?
The idea is to maintain a single event queue and handle most things with events (eg. a custom event "widget xy has requested a partial redraw at xx, yy"). A smart, application-centric event parser could then merge, skip, throttle and do other things with events, also being smart about partial redrawing, keeping pugl smartness at an easy level. Any thoughts?

osch:
> Then how do I know whether to draw the entire region that expose specifies, or only the ones that came before?

With `PUGL_DONT_MERGE_RECTS` set the redraw region is the union of all rectangles that was passed in exposure events with `count > 0` including the last one that is passed with `count == 0`.

> What happens when the window system sends exposes too?

Everything works, I see no problems here.

> What if they overlap?

Normally that doesn't matter: the application just has to draw the area consisting of the union of the passed rectangles. If the application needs to have rectangles that don't overlap than the application can split them but IMHO in most cases an application using drawing model 2 just needs an area and then tests which widgets intersect with the area and redraws only those widgets, thus an application using this mode is drawing widget oriented and not rectangle orientated, the rectangles are only for setting up a clipping region and for testing which widget has to be drawn.

> What is the advantage of all this complexity from the point of view of the application developer?

The advantage is that for applications with a lot widgets drawing performance is improved:
  - The advantage of having separate rectangles and not a unified rectangle is that lesser widgets are to be redrawn. 
  - The advantage of drawing the rectangles all at once and not each seperately is that some drawing operations (e.g. background widgets that are intersecting with more than one rectangle) have only to be done once for the whole list.

drobilla:
No, I mean the advantage of this convoluted expose model you are proposing, not having fine grained exposures at all.  Why would I want this? Why wouldn't I just want the library to tell me exactly what to draw at the right time?

osch:
Do you mean: why in my implementation the rectangles for the whole redraw area are given separately in different events and not at once as a whole list? There is no advantage in doing this, it was just easier to fit into the existing code base. It could also be easliy implemented to hold the list of rectangles in Pugl and give it at once to the application (at the point when my current implementation passes the event with count == 0)

drobilla:
Well, sort of.  I mean the concept of a deferred expose, then a later non-deferred expose that means you need to actually expose the earlier ones, etc.  They don't need to be provided literally as a list (events are POD anyway).

> There is no advantage in doing this, it was just easier to fit into the existing code base.

This is my point, really.  It significantly complicates the API, and adds a bunch of weird consequences (like Pugl having to send two exposes for a simple whole window refresh) for no benefit compared to Pugl doing it properly.

In case it's not blindingly obvious at this point, I'm not merging that kind of thing.  I am specifically working to get Pugl away from its history as a heavily forked collection of half-baked hacks that sometimes worked somewhere.  The quality of master (and the upcoming release) goes only up from here, never down.  I don't think even the current implementation is quite right, especially on MacOS...

> It could also be easliy implemented to hold the list of rectangles in Pugl and give it at once to the application.

Possibly.  If it's easy enough to do, then Pugl can do it.  If it's not, then it's not reasonable to expect applications to do it.  We'll have to see...

osch:
>  Pugl having to send two exposes for a simple whole window refresh

How did you come to this conclusion? This is not the case: for a simple window refresh there is only one expose event. Your answer gives me the impression that it could be that my previous explanations were misleading :-(

> I'm not merging that kind of thing

Yes I already assumed this, therefore I just made the implementation  minimal such that there are only very minimal changes to original Pugl as possible. Therefore I didn't introduce a rectangle list that is given to the application I reused the existing count variable in the expose event.

>  don't think even the current implementation is quite right, especially on MacOS

Why do you think so? I tested it thoroughly under Mac OS 10.13, 10.14 and 10.15 and if you have any hints that there could be some flaws it would be nice to give more details.

drobilla:
> This is not the case: for a simple window refresh there is only one expose event.

How is this expose any different from the one that comes after a bunch of deferred exposes then?

> don't think even the current implementation is quite right, especially on MacOS

See the tests branch.  I managed to get this to work there, but MacOS is weird because events are _dispatched_ in `puglPollEvents()`.  That's not how things are supposed to work, and it makes it impossible to control when events are dispatched in applications.  I'm not sure if it's actually possible to have separate wait and process functions there.  Maybe this was a mistake, it seems common, but I don't know if/why anyone would really miss it over a unified iterate function, which would also make some of the X11 cases a lot easier to deal with...

osch:
> How is this expose any different from the one that comes after a bunch of deferred exposes then?

Why should it be different? A simple window refresh will lead to just one expose event in the application with `count == 0` and full width and height at the end of the dispatch loop just as in normal Pugl. If there are more than one expose events while running the event loop (it doesn't matter if these are from the system or coming directly from the application) these events will be delivered with `count > 0` to the application and only the last exposure rectangle in the event loop is delivered with `count == 0`.

> I am not suggesting a rectangle list that is given to the application. I am saying the API doesn't need to change at all. Expose events as they currently are are fine.

This is only fine if the application uses drawing the simple model 1, but not for applications that are drawing widget oriented. If for example an application has two small widgets that are needing a refresh at the top left and bottom right corner of the window the application will get a merged rectangle for the whole window in the current implementation. This is not wanted for drawing model 2. In my implementation there is a much smaller area to be redrawn, since the two small rectangles are delivered as rectangle list to the application. This does work under Win, X11 and Mac < 10.14.


> MacOS is weird because events are dispatched in puglPollEvents()
> it makes it impossible to control when events are dispatched in applications

Yes, I also observed this when I implemented process timer in my LPugl branch (https://github.com/osch/lua-lpugl/blob/master/doc/README.md#world_setProcessFunc). It's nearly impossible to have full control when events will be dispatched under Windows and Mac OS and I wouldn't even try to do so. I don't see a unsolvable problem here, but I agree that the separate `puglPollEvents` and `puglDispatchEvents` are misleading since they only fit exactly to the x11 implementation. However it is also nice to have these functions separate for the x11 case.

> Also I suspect the clipping is still wrong there.

Ah sorry, I thought it was obvious: my current implementation requires the clipping-less drawing backend (with alpha layer drawing). As you don't want to have this, the implementation must be extended to provide all rectangles to the backend such that the backend can set up the clipping correctly. Again: my implementation just tries to have only minimal changes to original Pugl, since since I'm assuming that my modifications will not be merged. My current implementation was never meant for directly merging, it is a working and correct implementation and I only showed it to you as a proof of concept and perhaps starting point for improving Pugl to support drawing model 2 applications.

drobilla:
> A simple window refresh will lead to just one expose event in the application with count == 0 and full width and height at the end of the dispatch loop just as in normal Pugl.

> only the last exposure rectangle in the event loop is delivered with count == 0

So in one case, I get an expose with count 0 and I have to draw the whole thing.
In the other case, I get an expose with count 0 and I do not draw the whole thing.

Again, how do I know the difference?

7890:
> these events will be delivered with count > 0 to the application and only the last exposure rectangle in the event loop is delivered with count == 0

An old-school, full-rectangle-redraw app would check if count==0 for expose event, is that correct @osch ?

7890:
drobilla, what's the problem? wanna block me?

7890:
drobilla, why do you delete my comments?

drobilla:
@7890 This thread is bad enough without you running it off the rails.   Please stop.

osch:
> So in one case, I get an expose with count 0 and I have to draw the whole thing.
> In the other case, I get an expose with count 0 and I do not draw the whole thing.
> Again, how do I know the difference?

You always have to draw the area that is given by the exposure rectangle(s). If the exposure rectangle(s) cover the whole view than it's a full redraw. But a widget oriented application will not distinguish these cases: it will just consider the widget and redraws all widgets that are intersecting the area, and if the area is the whole view, than it automatically is a full redraw.

drobilla:
There is nothing special about a "full redraw".  The window system can expose part of the window too.  You can't use the area to distinguish these cases.

PR #23, opened 2019-09-11
X11 event handling fixes

osch:
Here are two fixes for X11 event handling.

1. pending redisplay requests are also send in `puglPollEvents`

2. `puglPollEvents` returns if there are events already read from the fd in the queue and does not wait until more events could be read from fd.

### DIFF
--- a/pugl/detail/x11.c
+++ b/pugl/detail/x11.c
@@ -106,31 +106,6 @@ puglInitViewInternals(void)
 	return (PuglInternals*)calloc(1, sizeof(PuglInternals));
 }
 
-PuglStatus
-puglPollEvents(PuglWorld* world, const double timeout)
-{
-	XFlush(world->impl->display);
-
-	const int fd   = ConnectionNumber(world->impl->display);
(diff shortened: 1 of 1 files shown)


pugl/detail/x11.c
```diff
@@ -597,6 +597,9 @@ puglPollEvents(PuglWorld* world, const double timeout)
 	sendAllRedisplayEvents(world);
 	XFlush(world->impl->display);
 
+	if (XEventsQueued(world->impl->display, QueuedAlready)) {
```
drobilla: When would this happen?

osch: Perhaps XFlush does also read some events from the fd itself into internal buffer? There are also other xlib function calls that are reading from the x11 connection. It's not specified which xlib function does some reading from the fd, so it's always a good idea to check if X11 read something before doinng a `select' on the fd.

drobilla:
Fair enough I guess.  Anyway, in general this seems good, I will clean it up and merge it.

drobilla:
077b318

PR #29, opened 2020-03-10
Added basic template for timer support, and windows implementation

check080:
After this commit, timers are working, documented and tested on windows/cmake/c++.

!! WARNING !! This may cause the mac or linux version to fail to compile, but it will likely be a small and fixable error.

Timers are also not yet implemented on mac or linux, since I haven't been able to test either one yet. I may be able to add linux timers myself, but I will need external help for mac timers.

### DIFF
--- a/pugl/detail/mac.m
+++ b/pugl/detail/mac.m
@@ -916,6 +916,19 @@ puglPollEvents(PuglWorld* world, const double timeout)
 	return PUGL_SUCCESS;
 }
 
+// TODO: see pugl.h for specification
+PuglStatus
+puglRegisterTimer(PuglView* view, uint64_t id, double rate){
+	return PUGL_FAILURE;
+}
+
+// TODO: see pugl.h for specification
+void
+puglDeregisterTimer(PuglView* view, uint64_t id){
+	
+}
+
+
 #ifndef PUGL_DISABLE_DEPRECATED
 PuglStatus
 puglWaitForEvent(PuglView* view)
--- a/pugl/detail/win.c
+++ b/pugl/detail/win.c
@@ -47,10 +47,11 @@
 #    define GWLP_USERDATA (-21)
 #endif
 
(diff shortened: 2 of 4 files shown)


drobilla:
Great, thanks.

I'm keen on having finally added tests to pugl recently, and this seems like an easy one, so I'll try to write one and implement it for the other platforms based on that.

Would probably be good to use in a demo too though.  Maybe one of them can rotate the cube or whatever every second or so when not running continuously...

drobilla:
This is way harder on both MacOS and especially X11.

It might be easier if pugl allocates the ID instead, but even then things don't quite line up...

check080:
Would it be easier if the function template looked more like:

PuglTimer* registerTimer(PuglView* view, double rate)

I could fix up the windows version to do that, if it helps

drobilla:
Don't bother, it's going to need to be massaged along with the implementations, but thanks.

A pugl-allocated uintptr_t ID will probably work for MacOS.  X11 has an xsync extension, but it doesn't allow associating any extra data with timers, and does not send events to windows at all, so... still no idea what to do there.

I imagine that reworking the event loop functions is going to have to happen for this.

drobilla:
I ended up doing this from scratch but after a few misadventures it turned out basically the same in the end anyway: a54361853bdfa08437c2858e603ce6202fb341b2

It wasn't too much of a nuisance to do the ID thing like the Windows API, since both X11 and MacOS (annoyingly) required storing a record of all the timers anyway.

It reliably gets about 2ms precision on X11 and MacOS (really more like the ideal 1, but these things are never perfect), but only 10ms on Windows which is the minimum period there.  There's some other "multimedia timers" API there that can apparently do better, but it's a whole different thing that doesn't work in the usual way and I don't care enough to bother.

We'll see how common functioning XSync extension support actually is, I have no idea...

PR #33, opened 2020-05-27
Remove macOS autorelease pool

jpcima:
Hello. Every time, on macOS, I will get a similar crash report when closing the Carla window which hosts my UI (VST3).

It's the draining of `autoreleasePool`, when destroying the Pugl world.
After deleting the releasepool code from Pugl, the problem stops happening.

Does it have any adverse effect to delete it?

Note: I tried debugging the program for memory errors, be it malloc gard / asan, in search of what may be another origin of this problem. It didn't detect any, all I ever saw is this backtrace of NSAutoreleasePool.

Perhaps it's my plugin-side Mac code which get mixed with this autorelease thing, I've no idea.

```
objc[77530]: autorelease pool page 0x7f9bbf95e000 corrupted
  magic     0x00000000 0x00000000 0x00000000 0x00000000
  should be 0xa1a1a1a1 0x4f545541 0x454c4552 0x21455341
  pthread   0x10e6435c0
  should be 0x10e6435c0

Abort trap: 6
```

### DIFF
--- a/pugl/detail/mac.h
+++ b/pugl/detail/mac.h
@@ -54,7 +54,6 @@
 
 struct PuglWorldInternalsImpl {
 	NSApplication*     app;
-	NSAutoreleasePool* autoreleasePool;
 };
 
 struct PuglInternalsImpl {
--- a/pugl/detail/mac.m
+++ b/pugl/detail/mac.m
@@ -776,7 +776,6 @@ puglInitWorldInternals(PuglWorldType PUGL_UNUSED(type),
 		1, sizeof(PuglWorldInternals));
 
 	impl->app             = [NSApplication sharedApplication];
-	impl->autoreleasePool = [NSAutoreleasePool new];
 
 	return impl;
 }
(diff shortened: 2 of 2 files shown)


drobilla:
TBH I don't understand this stuff very well, but I don't think removing it completely is okay, because then programs would have no autoreleasepool at all.  Maybe just don't create one if the world type is `PUGL_MODULE`?

jpcima:
Perhaps, it's not the correct idea having one that corresponds to the whole lifecycle of the world.
Examining vstgui for comparison, all observable uses of auto-release pools are observed only at function scope.

- the AutoreleasePool wrapper
https://github.com/steinbergmedia/vstgui/blob/develop/vstgui/lib/platform/mac/cocoa/autoreleasepool.mm

- the typical site of use
https://github.com/steinbergmedia/vstgui/blob/f326bb4853f4cc30283cdb5377a7e996c7df1df0/vstgui/lib/platform/mac/cocoa/nsviewframe.mm#L730

drobilla:
2d907d2 should fix the issue, but I still have no idea what the proper thing to do here is...

PR #34, opened 2020-05-28
win: SwapBuffers with OpenGL only

jpcima:
Using the cairo backend on Windows (wine), `SwapBuffers` would crash the program.

Most likely, this API relates to OpenGL only.
MSDN classifies it under OpenGL along with all of wgl API.
https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-swapbuffers

Not verified on real Windows but I suppose the fix to make sense for this case.


drobilla:
Seem to still work fine.  Merged as 21d8b08, thanks.

PR #35, opened 2020-05-31
Add editorconfig settings

jpcima:
This adds automatic indentation settings for editors.


drobilla:
I'd never even heard of this thing before, but sure, why not...

Merged as eb85b35, thanks.

PR #36, opened 2020-06-01
Add cursor change functionality

jpcima:
This work adds an ability to change the mouse cursor.
For now, implemented on X11 and Mac.

The cursor list is not exhaustive, it's based on the Windows set.
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursora

### DIFF
--- a/examples/pugl_cursors_demo.c
+++ b/examples/pugl_cursors_demo.c
@@ -1,0 +1,206 @@
+/*
+  Copyright 2012-2020 David Robillard <d@drobilla.net>
+
+  Permission to use, copy, modify, and/or distribute this software for any
+  purpose with or without fee is hereby granted, provided that the above
+  copyright notice and this permission notice appear in all copies.
+
+  THIS SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+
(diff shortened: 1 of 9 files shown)


jpcima:
Added win32 also, tested in wine.

drobilla:
Great, thanks.  I'll check it out.

How did you test it?  It needs to be added to one of the test programs, if you happen to already have that laying around anyway.

jpcima:
No test program. Testing was as follows.
- edited the cairo sample
- invoke random cursor change in two contexts, once before view realization, and on keypress event.
- ensure key press to have consistent behavior, depending if cursor is over the view during a press or outside

drobilla:
Nice demo!  Thanks.

It seems to get stuck sometimes on the wrong cursor for me (on X11), or sort of seems like it hangs for a bit, or... something.  I'll look into it this weekend.

drobilla:
How attached are you to the various cursors?  I'm not super comfortable with some of them not being distinct on X11 (the demo was confusing) and the use of private APIs on MacOS (particularly considering how Apple is about compatibility).  So although it's a bit annoying that e.g. there's no grab or universal resize available across them all, I think I'd feel better about only defining a common subset that is available on all three platforms.  As far as I can tell, this seems to be:

```
	PUGL_CURSOR_ARROW,
	PUGL_CURSOR_IBEAM,
	PUGL_CURSOR_CROSS,
	PUGL_CURSOR_HAND,
	PUGL_CURSOR_NO,
	PUGL_CURSOR_HRESIZE,
	PUGL_CURSOR_VRESIZE,
```

Which also happens to be what GLFW does, presumably for the same reasons.

drobilla:
See https://github.com/lv2/pugl/tree/cursors

jpcima:
This subset is not too bad for a start.

For the cursor mapping, I took it identical to a subset of Gdk, which itself contains duplicates, with exception of Cocoa undocumented cursors which are extra.

In Gdk, it identifies the cursors by strings, some of which are common between platforms, and some are only valid for specific platforms.

drobilla:
Fair enough.  I'm kind of torn, because some of the omissions are pretty annoying, but I think it's more in-line with the general philosophy of pugl to provide only a strict subset that works (roughly) the same everywhere.  The private Cocoa cursors are a bit of a curve-ball though... maybe it's okay...

I'll check what the common subset is when including those.  I'm not a big fan of things being duplicates on some platforms, it just feels broken to me, but if the subset gets better by using those APIs maybe that's the best compromise.

(Alternatively I could be lazy and just merge this for now because we can always add more later anyway)

drobilla:
Merged as 2f90b71 with just the common subset for now, thanks.  Seems like there's not much except a wait cursor that's really common, maybe a "general drag around" if you get a bit fuzzy, but I'll just commit the ultra conservative thing for now.  Feel free to make another PR if you want.

PR #37, opened 2020-06-15
Minor cursor tweaks for Mac and Windows

7890:
On Mac:
Handle the following case:
-Set mouse cursor when window got focus

Test:
-Start pugl_cursor_demo
-Click outside of window
-Click inside window -> cursor doesn't change without this patch, only after leaving / entering with the mouse while window has focus.

Side-topic: should PUGL_POINTER_IN / OUT be emitted when window got focus via mouse click?

On Windows:
The cursor set by pugl also appears on window borders, minimize / maximize buttons etc.
Forwarding WM_SETCURSOR if it wasn't set by pugl will show standard resize etc. cursors on window borders.

Test: start pugl program, move mouse towards and over the window border.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,3 +10,4 @@ Jordan Halase <jordan@halase.me>
 Oliver Schmidt <oliver@luced.de>
 Zoë Sparks <zoe@milky.flowers>
 Jean Pierre Cimalando <jp-dev@inbox.ru>
+Thomas Brand <tom@trellis.ch>
--- a/pugl/detail/mac.m
+++ b/pugl/detail/mac.m
@@ -354,6 +354,11 @@ handleCrossing(PuglWrapperView* view, NSEvent* event, const PuglEventType type)
 	puglview->impl->mouseTracked = false;
 }
 
+- (void) cursorUpdate:(NSEvent*)event
+{
+	[puglview->impl->cursor set];
+}
+
 - (void) mouseMoved:(NSEvent*)event
 {
 	const NSPoint         wloc = [self eventLocation:event];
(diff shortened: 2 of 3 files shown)


drobilla:
I can't figure out any observable effect of the first commit.

7890:
If a cursor is set and then at some point focus is lost .. then focus comes back eg. click inside window -> it seems currently the cursor set before focus lost is not the one getting back. There's a difference when the mouse enters the window by crossing the window edge, it correctly switches the cursor. When clicking inside the window from another focus, the described pattern happens. I will try again -

7890:
```
- (void) cursorUpdate:(NSEvent*)event
{
	[puglview->impl->cursor set];
}
```
is analogous to
```
	case WM_SETCURSOR:
		if (LOWORD(lParam) == HTCLIENT) {
			SetCursor(view->impl->cursor);
		}
```

7890:
For a clean build and install of d70aa80d71299c24ebd71537cfd620eff2b7e50d, I get this exact behaviour on OSX `sw_vers -productVersion` 10.13.6:

in pugl/build:
`open pugl_cursor_demo.app`

- [x] move mouse to second quad in first row -> mosue cursor changes to caret 

- [x] focus anything (other than 'Pugl Window Demo') -> mouse cursor changes to anything

- [ ] click-focus 'Pugl Window Demo' with the mouse inside the second quad in first row (expect caret cursor) -> cursor does not change. << potential issue here.

- [x] move mouse inside now focused window -> cursor updates to caret

with patch:

- [x] cursor updates to caret on mouse focus-click button up (right after the button down/up sequence), without a mouse movement necessary to get the expected cursor

Please verify.

drobilla:
I can reproduce that issue but the change does not fix it.

drobilla:
Merged as 743db99 and db0a48e, thanks.

PR #41, opened 2021-01-24
Cast int to double [-Wnarrowing]

sjaehn:
Only to get rid of compiler warnings.

### DIFF
--- a/src/x11.c
+++ b/src/x11.c
@@ -1120,7 +1120,10 @@ puglDispatchX11Events(PuglWorld* world)
       XGetWindowAttributes(view->impl->display, view->impl->win, &attrs);
 
       const PuglEventConfigure configure = {
-        PUGL_CONFIGURE, 0, attrs.x, attrs.y, attrs.width, attrs.height};
+        PUGL_CONFIGURE, 0, 
+        (double)attrs.x, (double)attrs.y, 
+        (double)attrs.width, (double)attrs.height
+      };
 
       puglDispatchEvent(view, (const PuglEvent*)&configure);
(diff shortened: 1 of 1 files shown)


drobilla:
How are you building to trigger this?  With g++?

sjaehn:
Right, g++.

drobilla:
Why?

(This change is fine, and I imagine it works because of the MSVC build which is in C++, but I don't actually test this and wonder why you'd build a C library with g++ / if it should be covered / etc)

sjaehn:
Ping! I got your point. Pugl should now be build as a library, only. No backpacking anymore (you told you wanted to get rid of this). Then you are right. g++ compiling doesn't make any sense.

However, I still use pugl in the "old way". Backpacked to my projects. Like the most pugl using developers still do.  We had a discussion about this on last weekend on irc#lad (in your absence).

drobilla:
I see.  That's fine, you can link C and C++ objects together without issue, even if Pugl is vendored.

Whatever I suppose, since Pugl happens to be written in the subset of C and C++ this is fine, but in general it's strange.  C is not a subset of C++, and most C code is not capable of this.

drobilla:
FYI your name and email address is not configured in git.  Do you really want "42939779+sjaehn@users.noreply.github.com" in git history?

In this case it's trivial so whatever, but in general I don't think I'm comfortable merging commits without proper attribution.

drobilla:
Effectively equivalent fix committed as e497eea68ca704f62d3a65f1bbe787167cd3ae3e, thanks.

PR #44, opened 2021-04-22
Prevent memory leak on MacOS

bjarkebech:
When testing an OpenGL view on my system, the many NSEvents and NSDates fired on each ```puglUpdate``` call kept being allocated new memory, making the RAM usage increase over time.

Wrapping ```puglUpdate``` in an @autoreleasepool fixed it for me.

Also – absolutely love the library. Exactly what I was looking for. 😄

### DIFF
--- a/src/mac.m
+++ b/src/mac.m
@@ -1188,36 +1188,38 @@ dispatchClientEvent(PuglWorld* world, NSEvent* ev)
 PuglStatus
 puglUpdate(PuglWorld* world, const double timeout)
 {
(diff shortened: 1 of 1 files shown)


drobilla:
Ah, yeah.  Still a bit of a mystery to me how that stuff works, to be honest.  I imagine there's other places where it's a bit leaky...

Anyway, this seems reasonable.  Merged as e67425c7f32d4ea25dfb92e44a9331914ae3d47e, thanks.

PR #45, opened 2021-05-23
Fix the declaration of WGL function types

jpcima:
Fixes some crashes in 32-bit OS


drobilla:
I didn't try on 32-bit, but everything still works fine for me on 64 anyway.  Merged as 922057c, thanks.

PR #46, opened 2021-05-25
Fix narrowing compiler warnings

falkTX:
Pretty straight forward.
The macOS one is very odd that the compiler was complaining about converting integers, my guess is that NSOpenGLPixelFormatAttribute is less than 32bits in size.

### DIFF
--- a/src/mac_gl.m
+++ b/src/mac_gl.m
@@ -73,8 +73,8 @@
     NSOpenGLPFAColorSize,     colorSize,
     NSOpenGLPFADepthSize,     (unsigned)puglview->hints[PUGL_DEPTH_BITS],
     NSOpenGLPFAStencilSize,   (unsigned)puglview->hints[PUGL_STENCIL_BITS],
-    NSOpenGLPFAMultisample,   samples ? 1 : 0,
-    NSOpenGLPFASampleBuffers, samples ? 1 : 0,
+    NSOpenGLPFAMultisample,   (NSOpenGLPixelFormatAttribute)(samples ? 1 : 0),
+    NSOpenGLPFASampleBuffers, (NSOpenGLPixelFormatAttribute)(samples ? 1 : 0),
     NSOpenGLPFASamples,       samples,
     0};
   // clang-format on
--- a/src/win.c
+++ b/src/win.c
@@ -451,7 +451,7 @@ initKeyEvent(PuglEventKey* event,
 static void
 initCharEvent(PuglEvent* event, PuglView* view, WPARAM wParam, LPARAM lParam)
 {
-  const wchar_t utf16[2] = {wParam & 0xFFFF,
+  const wchar_t utf16[2] = {(wchar_t)(wParam & 0xFFFF),
                             (wchar_t)((wParam >> 16) & 0xFFFF)};
 
   initKeyEvent(&event->key, view, true, wParam, lParam);


falkTX:
> Which warning flags on which compilers triggered them?

Building in C++ mode :)

drobilla:
Ah.  Meson annoyingly can't do that (as far as I can tell) except with MSVC since you can pass `/TP` and ignore what the build system thinks.  I started building as C on MSVC since it works on recent enough versions, but it's annoying to not have tests for building as C++.

That said... why would you do this with clang/GCC on Mac/Linux?

falkTX:
I am namespacing the whole pugl (except on macOS, because objc doesnt allow it).
Reason for this is avoid symbol clashes as much as possible, even with debug builds.
If a plugin has PluginA::puglSetWindowTitle and another one has PluginB::puglSetWindowTitle they will not conflict even with full on debug mode and symbols.

drobilla:
I don't think trying to jam every C library into a C++ namespace is a reasonable thing to do.  C plugins (or any plugin that uses a C library) would have this problem anyway.  Fix your build system?

falkTX:
Unreasonable maybe, but sometimes practical.
Making a monolithic file to build an entire API is very handy, as shown by the many single-header-files implementations like stb.

The problem is not building the files on my case, but how I wish to setup things.

In any case, these are just warnings, it doesnt prevent building. Can be easily ignored or muted.
I think the windows one also triggered under cygwin, will check for that at a later point.

drobilla:
Sure, fixing warnings are fine.  I'm more concerned about the long-term repercussions of mangling the code.

> Making a monolithic file to build an entire API is very handy, as shown by the many single-header-files implementations like stb.

Maybe, but pugl is not such a library, and many (most) libraries aren't either.  You just can't arbitrarily concatenate whatever code together and expect it to work in general (and this exact thing has been a problem before...).  Compilation units *are* namespaces.  This is essentially the C equivalent of going to a C++ library, removing all namespaces, and expecting it to work.  Which of course it won't, because the code was written in those namespaces.  It's not feasible for me to deal with a never-ending series of "so I removed all the namespaces from your code for no particular reason and it broke again".

Anyway, my logic WRT the C++ namespace is this: if this is a problem for C plugins, then it needs to be fixed.  If it's not, then there's no reason to put this C library in a C++ namespace anyway.  In other words, this only causes problems, and at best avoids an instance of one you have anyway, since plugins can and will have all kinds of private clashing symbols.  You can't put the entire world in a custom namespace.

That said, if you really, really, really, really want pugl to build as a single compilation unit for some good concrete reason, make a PR that builds and tests it that way.  Then it won't rot and I'll consider it a supported deployment style.  Current it is not, and if by some coincidental miracle it happens builds that way at the moment, that will definitely break repeatedly in the future.  I don't see any real reason to do this (it won't magically fix clashing symbol issues), since this won't fix any clashing symbol issues.  That said, that seems like a waste of time to me, and I think you should try to fix your issues by defining `PUGL_API` appropriately.

drobilla:
Warnings (hopefully) fixed, if slightly differently, in 040e94b.  Thanks.

falkTX:
Can confirm warnings are gone for windows case, thanks.

and final note..
I understand the problems with this approach. if it becomes problematic I will find out alternatives, or just go the regular way.
I am fine with dealing with potential issues for this myself.
(and btw, the entire plugin stuff is inside a namespace, separate from the GUI toolkit thingies. this namespace is easily changed by a macro)

The addition of a test case that verifies that this whole thing is a nice idea.
I think I will try that. But dont be too grossed out by the result.

drobilla:
It somehow never occurred to me that both building as C++ and building all in one compilation unit is relatively easy to test at once by writing a C++ test file that includes them all.  It's a bit annoying compared to being able to use the build system, and doesn't really scale to being an approach to use for all the other libraries, but it's not so bad for pugl anyway.

So: 9ecda14

falkTX:
That is very welcome, thank you very much!

PR #47, opened 2021-05-25
Make win32 puglSetFrame consistent with x11

falkTX:
The x11 code only sets frame if `XMoveResizeWindow` succeeds.
This makes the win32 code behave the same way, it will only set the frame if `SetWindowPos` succeeds.


drobilla:
Merged as a567490, thanks.

PR #48, opened 2021-05-25
x11: do not override _POSIX_C_SOURCE if already defined

falkTX:
Yet another simple compiler warning fix.


drobilla:
Merged as ab80b4b, thanks.

PR #51, opened 2021-06-07
Fix a macOS warning

falkTX:
Just something tiny.

### DIFF
--- a/src/mac.m
+++ b/src/mac.m
@@ -1237,7 +1237,7 @@ puglUpdate(PuglWorld* world, const double timeout)
                                              untilDate:date
                                                 inMode:NSDefaultRunLoopMode
                                                dequeue:YES]);) {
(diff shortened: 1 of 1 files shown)


drobilla:
Merged as e52c00e, thanks.

Seems reasonable, but (obviously) I get no warning here.  For future PRs that fix warnings, please include the warning message triggered and compiler version in the commit message so I can verify them and/or update the build system to match.

falkTX:
Will do. This time I saw the warning and commit/pushed only later on non-macOS machine, so didnt have the specific output at hand.

Anyway, the warning I get is:
```
src/pugl-upstream/src/mac.m:1240:61: warning: comparison of different enumeration types ('NSEventSubtype' and 'PuglEventType') [-Wenum-compare]
      if ([ev type] == NSApplicationDefined && [ev subtype] == PUGL_CLIENT) {
```

This is on both macOS 10.14 and 11.2, maybe another C++ specific one, not sure.

PR #52, opened 2021-06-07
Clear macOS GL context after creating window and initial size

falkTX:
Run into issues when using multiple GL UIs with pugl.
Opening new GL UI can mess up the state of existing UIs.
This was partly at fault.


drobilla:
I suspect there are other problems if this fixed anything, but resetting the GL context is a good idea anyway.  Merged as c98ee49, thanks.

PR #55, opened 2021-07-02
Only call SetProcessDPIAware if program type

falkTX:
Fixes plugins changing global DPI host state
(which makes the whole application unusable)

See https://github.com/lv2/pugl/issues/54 for the related issue.


drobilla:
Merged as fbf5ed0b, thanks.

Issue #56, opened 2021-07-26
Nested event loop crash


reuk:
Potentially-linked issue in DPF: https://github.com/DISTRHO/DPF/issues/299

It looks like when LV2 UIs call `puglUpdate` from the idle callback on macOS, they can occasionally process events which cause the UI to be destroyed. This in turn causes the host to crash, because the next part of puglUpdate tries to use resources which have been freed during `sendEvent`.

The issue linked above contains some steps to reproduce the issue.

I believe that it is an error to call `NSApplication sendEvent` from within a plugin. It places a very large burden on the host to assume that any event could be processed inside an LV2 idle callback. Hosts which weren't designed with LV2 in mind may have to undergo significant rewrites in order to work around this issue, if they can at all. It seems much more straightforward to provide both "standalone" and "plugin" versions of puglUpdate, so that users of pugl can avoid manually processing events in contexts which already have an event loop in place.

Would it be possible to modify pugl such that `sendEvent` can be bypassed in projects with an existing event loop?

falkTX:
pugl knows during runtime if the program is standalone or a plugin, so it can make a decision based on that.
A similar issue was pointed out in https://github.com/lv2/pugl/pull/55 where plugins are calling a global function typically reserved for hosts/standalones.

drobilla:
The first part sounds like a bug that just needs to be fixed regardless, but I don't know about the second part.  The sendEvent is there to "put back" events that pugl is not going to process because the event mask is not precise enough to filter out exactly the events that are handled specially.

The intent is not to "process" events there, just the opposite: sending the event to the application is the way pugl "puts back" the event, for lack of a proper way to do that (that I can find anyway).  But MacOS is very unpredictable internally in general, IIRC nextEventMatchingMask itself dispatches events sometimes, and other system things do as well, more or less whenever.  I am not sure this is possible to entirely avoid, even if pugl doesn't call sendEvent.

It's also not clear to me why this is problematic anyway.  A more concrete example would help.  But anyway, in short...

> Would it be possible to modify pugl such that sendEvent can be bypassed in projects with an existing event loop?

I have no idea.

Maybe it's possible to specify just NSApplicationDefined as a mask to handle the client events, but this would only reduce the scope of the problem (though by quite a bit), or specify false for `dequeue`, but my experience with this method is that it's very cryptic and temperamental and usually doesn't do what you expect, so it will probably take some tinkering.

[...]

PR #57, opened 2021-07-30
macOS event loop fixes

reuk:
When running pugl as a PUGL_MODULE, it's dangerous to call `sendEvent` on arbitrary events, as processing such events could cause the UI instance to be destroyed. Then, when statements later on in `puglUpdate` are executed, they may attempt to access resources that have been freed, crashing the process. This is currently seen in LV2 UIs which execute puglUpdate from the idle callback.

This change just posts events to the CFRunLoop directly, which seems a bit cleaner. This way, there's no need to convert the event to an NSEvent and back, as we can just capture the event in a block instead.

I also noticed that the test_redisplay app asserts on Big Sur because the OS invalidates the entire window in response to the invalidate request, so I've updated the test a little to be slightly more forgiving about mismatches between the requested and actual repainted areas.

Finally, I noticed that the meson build currently doesn't work out-of-the-box due to undefined references to `c_warnings` and `objc_warnings`. Perhaps this is a configuration issue on my system, so I haven't pushed a fix. That said, removing the references allowed the build to succeed.

See also #56 and https://github.com/DISTRHO/DPF/issues/299.

### DIFF
--- a/src/mac.m
+++ b/src/mac.m
@@ -1174,23 +1174,21 @@ PuglStatus
 puglSendEvent(PuglView* view, const PuglEvent* event)
 {
   if (event->type == PUGL_CLIENT) {
-    PuglWrapperView* wrapper = view->impl->wrapperView;
-    const NSWindow*  window  = [wrapper window];
-    const NSRect     rect    = [wrapper frame];
-    const NSPoint    center  = {NSMidX(rect), NSMidY(rect)};
+    PuglEvent copiedEvent = *event;
 
-    NSEvent* nsevent =
-      [NSEvent otherEventWithType:NSApplicationDefined
-                         location:center
-                    modifierFlags:0
-                        timestamp:[[NSProcessInfo processInfo] systemUptime]
-                     windowNumber:window.windowNumber
-                          context:nil
-                          subtype:PUGL_CLIENT
-                            data1:(NSInteger)event->client.data1
-                            data2:(NSInteger)event->client.data2];
+    CFRunLoopObserverRef observer = CFRunLoopObserverCreateWithHandler(
+      NULL,
+      kCFRunLoopAfterWaiting,
+      false,
+      0,
+      ^(CFRunLoopObserverRef, CFRunLoopActivity) {
+        puglDispatchEvent(view, &copiedEvent);
+      });
 
-    [view->world->impl->app postEvent:nsevent atStart:false];
+    CFRunLoopAddObserver(CFRunLoopGetMain(), observer, kCFRunLoopCommonModes);
+
+    CFRelease(observer);
+
     return PUGL_SUCCESS;
   }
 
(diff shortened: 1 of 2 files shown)


src/mac.m
```diff
@@ -1205,48 +1203,27 @@ - (void)windowDidResignKey:(NSNotification*)notification
...
-        dispatchClientEvent(world, ev);
-      }
-
-      [world->impl->app sendEvent:ev];
-
-      if (timeout < 0) {
-        // Now that we've waited and got an event, set the date to now to avoid
-        // looping forever
-        date = [NSDate date];
+    if (world->impl->autoreleasePool != nil) {
```
falkTX, line 1210: This check looks a little weird, might be worth mentioning that `autoreleasePool` is only non-null for `PUGL_PROGRAM` types.

falkTX: I mean mention as a code comment.

reuk: It might be nicer to introduce a `PuglWorldType puglWorldGetType (PuglWorld const*)` which could be used in this location. That would require implementation for other platforms too, though.

src/mac.m
```diff
@@ -1174,23 +1174,21 @@ - (void)windowDidResignKey:(NSNotification*)notification
...
-    [view->world->impl->app postEvent:nsevent atStart:false];
+    PuglEvent copiedEvent = *event;
+
+    CFRunLoopObserverRef observer = CFRunLoopObserverCreateWithHandler(
+      NULL,
+      kCFRunLoopAfterWaiting,
+      false,
+      0,
+      ^(CFRunLoopObserverRef, CFRunLoopActivity) {
+        puglDispatchEvent(view, &copiedEvent);
```
drobilla, line 1185: Isn't this referring to garbage on the stack (copiedEvent)?

src/mac.m
```diff
@@ -1174,23 +1174,21 @@ - (void)windowDidResignKey:(NSNotification*)notification
...
+      });
+
+    CFRunLoopAddObserver(CFRunLoopGetMain(), observer, kCFRunLoopCommonModes);
```
drobilla, line 1188: According to the documentation, this does nothing if there is already an observer registered for the same mode.  Doesn't this mean multiple client events won't work?  Does it make sense to add a run loop observer for every event anyway?

drobilla (commented):
Thanks.

This is some pretty black magic, but I'll try to verify it.  In the meantime I left a couple comments on things that seem questionable at first glance.

src/mac.m
reuk: Are you referring to this part of the docs?
> If rl already contains observer in mode, this function does nothing.

I interpreted that to mean that the function was a no-op if _that specific observer_ had already been added in the specified mode. Perhaps I can test this by loading a couple of DPF plugins in REAPER and checking whether client events get sent for both.

src/mac.m
reuk: It doesn't look that way in a debugger.

According to [this page](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html#//apple_ref/doc/uid/TP40011210-CH8-SW3), referenced objects are copied unless you specify otherwise.

src/mac.m
reuk: I've tested this out a bit in some modified DPF plugins. I added two calls to puglSendEvent inside puglUpdate (which I knew was guaranteed to be called regularly by the host), both with different client events. Then, I stuck a breakpoint inside the block passed to `CFRunLoopObserverCreateWithHandler`. In LLDB, I can see both events getting triggered, which indicates that adding the second observer is not a no-op, and does not remove the first observer.

I also tried loading up multiple plugin instances with the same modification, and checked that both received the client event callbacks. It appears that all plugin instances which register observers for a particular mode receive the callbacks that they register.

src/mac.m
drobilla: The guts of the world/everything are available anywhere in the implementation, so the accessor part isn't really necessary (though it doesn't hurt either).

src/mac.m
drobilla: Cool, thanks for looking into it.

The "timing" of this approach seems weird to me, since more or less the whole point of `puglUpdate` is to process all events _now_, but I'll get into my Mac at some point this weekend and tinker with it a bit.  It might not matter since MacOS is already wildly divergent from the other platforms WRT event timing/order semantics anyway...

test/test_redisplay.c
```diff
@@ -55,6 +55,22 @@ typedef struct {
 static const PuglRect  redisplayRect   = {2, 4, 8, 16};
 static const uintptr_t postRedisplayId = 42;
 
+static bool
+rectContains(PuglRect outer, PuglRect inner)
+{
+  return outer.x <= inner.x && outer.y <= inner.y &&
```
falkTX, line 61: this line is a bit confusing, because the `outer` comes first here, while below it is the `inner`.
I propose to change this to:
```
  return inner.x >= outer.x && inner.y >= outer.y &&
```

drobilla:
Well, there's a rebased version at https://github.com/lv2/pugl/tree/nested-event-loop-fix and it doesn't break the tests, but the tests don't currently cover modules.

I don't particularly understand this, but it still seems like an elaborate way to do "make client events not work in modules in MacOS" (and "make `puglUpdate` do nothing") to me.  If it's literally impossible to dispatch events within a specific scope on MacOS... okay I guess?  That seems unlikely though?

In any case, I don't have the ability to test this at the moment, but if any of you can, and if there's some reason why making `puglUpdate` work properly is impossible, I can probably be convinced to merge it.

falkTX:
if it is worth anything I have been using pugl with this patch/change applied for as long as it has been originally reported, and have seen no issues (contrary to before the patch).

drobilla:
@falkTX You snuck in before the edit, I hadn't noticed that @reuk said it has been tested.

Client events are only used for... well, code that explicitly uses client events, so pugl would generally work aside from that either way.

falkTX:
right I never make use of those for the moment.

drobilla:
Merged as 28d75f7, thanks.

I had to change the run loop mode to `kCFRunLoopBeforeSources` to get reliable behaviour for programs.  Hopefully that doesn't cause any new issues...

PR #58, opened 2021-08-18
RFC: Possible fix for macOS transient windows always visible

falkTX:
pugl supports transient parent windows, for which the parent is not something to embed but just something to place below a window on the "WM window list stack".

I have an issue when creating 2 views, one is a transient parent to the other.
It works fine everywhere except macOS.

on macOS, having `makeKeyAndOrderFront` will always make the window visible on screen.
Making the window key seems wrong in this case, the key window should be the transient parent, not the "child"/top window.

Comments welcome, I verified that this patch works correctly for my use case

### DIFF
--- a/src/mac.m
+++ b/src/mac.m
@@ -1036,10 +1036,14 @@ puglRealize(PuglView* view)
     puglSetFrame(view, view->frame);
 
     [window setContentView:impl->wrapperView];
-    [view->world->impl->app activateIgnoringOtherApps:YES];
     [window makeFirstResponder:impl->wrapperView];
-    [window makeKeyAndOrderFront:window];
-    [impl->window setIsVisible:NO];
+    [window setIsVisible:NO];
+
+    if (! view->transientParent)
+    {
+      [window makeKeyAndOrderFront:window];
+      [view->world->impl->app activateIgnoringOtherApps:YES];
+    }
   }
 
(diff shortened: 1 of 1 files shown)


falkTX:
To be clear: I have 2 windows which are created at the start of the program (well, plugin in this case).
The 2nd window is used as an "about this software" sorta thing, it is created but hidden until the user clicks on the brand/logo.

On Windows and X11 systems, this 2nd window is correctly not visible until the user clicks on said button.
On macOS this 2nd window starts visible (but without contents, as the NSView inside is in a hidden state). Pressing the brand button that would usually show the about-window makes the (already wrongly visible window) draw the contents as normal. After this we can close the about-window and things behave as expected again. It is just the initial state that is misbehaving/wrong.

drobilla:
Seems reasonable enough, thanks.  I'm away right now and without a Mac, but I'll test it out later this week when I get home.

falkTX:
I am closing the PR since pugl changed a fair bit, and this likely needs a bit of tweaking.
Likely partially or even completely fixed now anyway, still need to check this one.

drobilla:
Yes, although transient window stuff isn't really solid or tested.  I intend to add a demo and/or test for that so it can be finished up, and other things (like the position) stuff done properly (or opinionatedly removed, or whatever, in that case).

PR #60, opened 2021-08-27
Check for xrandr extension before usage

falkTX:
There might be a situation where xrandr is enabled in the build, but it is not provided/supported by the x11 server.
One such case is X11 forwarding over SSH.
On such setups `XRRGetScreenInfo` crashes the whole application.

This PR fixes the issue by checking if xrandr is available before attempting to use it.


drobilla:
Merged as 34de31a0 (just superficial changes), thanks.

PR #74, opened 2021-12-15
Add diagonal cursors

### LINKED ISSUE #73: Wrong default cursor under KDE
When doing nothing with the pointer (default, `puglSetCursor` never called) or setting it to `PUGL_CURSOR_ARROW` it looks incorrect under KDE.

Screenshot_20211215_101352
 
I have seen this before on some raw X11 applications, looks very ugly and out of place.
The cursor is supposed to look like this: (which I can still get by building pugl without `HAVE_XCURSOR` defined.

Screenshot_20211215_101653

falkTX:
This is a WIP changeset to:
1. add diagonal cursor types (done on X11, not yet on other platforms)
2. use X11 cursor names instead of predefined macros (which lack the diagonal cursors, among others)

Note: I had to remove the first `defineCursorShape(view, impl->cursorShape);` on the view creation as it leads to a crash.
The `cursorShape` (now renamed `cursorName`) is set to the default value, so it does not have any visible difference.

Fixes #73

I will try to implement the diagonal cursors for macOS and Windows soon.

@drobilla any opinions on this, should I do it some other way?

### DIFF
--- a/include/pugl/pugl.h
+++ b/include/pugl/pugl.h
@@ -1251,13 +1251,15 @@ puglPostRedisplayRect(PuglView* view, PuglRect rect);
    Windows.
 */
 typedef enum {
-  PUGL_CURSOR_ARROW,      ///< Default pointing arrow
-  PUGL_CURSOR_CARET,      ///< Caret (I-Beam) for text entry
-  PUGL_CURSOR_CROSSHAIR,  ///< Cross-hair
-  PUGL_CURSOR_HAND,       ///< Hand with a pointing finger
-  PUGL_CURSOR_NO,         ///< Operation not allowed
-  PUGL_CURSOR_LEFT_RIGHT, ///< Left/right arrow for horizontal resize
-  PUGL_CURSOR_UP_DOWN,    ///< Up/down arrow for vertical resize
+  PUGL_CURSOR_ARROW,         ///< Default pointing arrow
+  PUGL_CURSOR_CARET,         ///< Caret (I-Beam) for text entry
+  PUGL_CURSOR_CROSSHAIR,     ///< Cross-hair
+  PUGL_CURSOR_HAND,          ///< Hand with a pointing finger
+  PUGL_CURSOR_NO,            ///< Operation not allowed
+  PUGL_CURSOR_LEFT_RIGHT,    ///< Left/right arrow for horizontal resize
+  PUGL_CURSOR_UP_DOWN,       ///< Up/down arrow for vertical resize
+  PUGL_CURSOR_DIAGONAL,      ///< Top-left to bottom-right arrow for diagonal resize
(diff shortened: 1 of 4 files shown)


falkTX:
https://github.com/glfw/glfw/commit/7dbdd2e6a5f01d2a4b377a197618948617517b0e seems to have some nice little code we can use for inspiration :)

drobilla:
Those symbols on MacOS are undocumented and private implementation details.

falkTX:
ah, that is the one I didnt try yet.
a bummer, but maybe something we should allow? the corner resize handle is a very useful one, it was the very reason I just began testing cursor stuff. (VST2 plugins do not allow host-side resize, so plugins need to do that themselves)

maybe put the macOS side behind a "I know this is dangerous, let me use it anyway" compiler macro?

drobilla:
I dunno.  I figured if it isn't in the standard set of cursors on MacOS, it's probably not that important.  The diagonal thing does seem weird though, since they are clearly used there.

Judging by the glfw docs they are "maybe" supported on both X11 and Wayland as well.  Aside from the fragility of using undocumented APIs, this is the main thing I wanted to avoid (the real possibility of a cursor change just not working at all).  It's not the end of the world, I guess, but this is why the set is currently is what it is anyway.

> maybe put the macOS side behind a "I know this is dangerous, let me use it anyway" compiler macro?

Though it'd be nice in theory to provide a compile time option to avoid this scenario, it doesn't seem likely enough to be worth the hassle.

falkTX:
IMO the corner resize cursor is essential.
I will do the work to get them supported as best as possible, then you see if it makes sense in upstream pugl or not.

The glfw X11 "maybe" is due to not all DEs/WMs providing XDG-specified cursor themes.

But I think that is quite fair, if a cursor is not supported just show the default.
It would be the same as no cursor being set, which I think is quite okay.

drobilla:
I think it needs to return an appropriate error code in any case where it doesn't work (error handling in general is something I'm trying to improve).  I wouldn't deliberately set it to the default, that will just mask issues.

I don't know how the cursor theme stuff works at all.  The actually correct thing to do for "default" probably needs looking in to, I suspect that I originally made it explicitly set the "default" that seemed to be right on my machine, and you've done the same on your machine, and we're probably both wrong :)

falkTX:
FYI just updated this to track latest pugl. Since you added those cursor names on pugl side, the changes are quite minimal now.

All previous points about macOS not officially supporting this still valid, of course.

falkTX:
windows and macOS implementations added

drobilla:
Merged as cdfb822 , thanks.

I changed the enum names and factored out a bit of the MacOS implementation.  Otherwise the same.

drobilla:
Oh, and updated `pugl_cursor_demo`.

PR #83, opened 2022-06-03
Bad locale

pacew:
I encountered a SIGSEGV crash when running Aether in a docker container running the official archlinux image.  The immediate cause was XUnsetICFocus being called with an argument of NULL.

Upon investigation, I found that the creators of the archlinux image have aggressively pruned /usr/share/X11/locale such that XOpenIM fails and returns NULL.  This causes XCreateIC to also return NULL, and finally crash the program when a FocusOut event arrives.  (The FocusIn doesn't crash because XSetICFocus happens to be safe to call with NULL.)

I was able to get the Aether plugin to run by putting a check around the call to XUnsetICFocus in src/x11.c or by populating the missing files /usr/share/X11/locale/C and /usr/share/X11/locale/iso8859-1.

This PR attempts to be a more complete defense than just putting a check around XSetICFocus, but unfortunately, I don't have a good way to set up my test environment again.  I hope this will be useful nonetheless.

I will also try to contact the archlinux image builders and ask that they make /usr/share/X11/locale complete enough that XOpenIM will succeed.

Thanks for your library!

### DIFF
--- a/src/x11.c
+++ b/src/x11.c
@@ -460,13 +460,13 @@ puglRealize(PuglView* const view)
   // Create input context
   if (world->impl->xim) {
     impl->xic = XCreateIC(world->impl->xim,
-			  XNInputStyle,
-			  XIMPreeditNothing | XIMStatusNothing,
-			  XNClientWindow,
-			  impl->win,
-			  XNFocusWindow,
-			  impl->win,
-			  (XIM)0);
+                          XNInputStyle,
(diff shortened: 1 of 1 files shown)


drobilla:
LGTM, thanks.  No worries if this isn't complete resilience to weird locales or input configuration, further issues can be dealt with when they crop up.

Merged as 2e0fb50d, thanks.

Issue #98, opened 2023-02-15
Display offset for x11/cairo in plugins


sjaehn:
I observed that the visual content written to the pugl-provided Cairo context is displayed with an offset in Carla, QTractor, Zrythm, and Reaper. And the content can't be moved (at least I don't see a way). So all the time parts of the plugin content are clipped. The unwanted offset is sometimes positive, sometimes negative. Here an example with the current master of B.Angr (build with the current pugl) in Carla:
![Screenshot B.Angr in Carla](https://user-images.githubusercontent.com/42939779/218993054-f6dfe2e9-f3f6-400a-b45e-c3ca48cd8ff1.png)

As there's an offset between the host-provided window area and the plugin-used area, the area not used by the plugin will become full of glitches with time (here: bottom and right). Especially on window resize. Likely as the plugin can't write on this part (to clean it).

Additional information:
* Observed in the latest pugl commits (also before [Fix partial exposure in X11 Cairo backend](https://github.com/lv2/pugl/commit/22988036924c706860d6155e5f728ff927ecca5b)).
* **NOT observed in old pugl builds (Jan 2021) I used before - Therefore I think it's pugl-related.**
* Independent from deprecated:resize. If I remove all resize calls (feature, extension) and definitions (in ttl), the problem still persists. Tried out with the simple B.Amp.
* **This problem does NOT occur in jalv and Ardour.**

sjaehn:
I dumped `view->lastConfigure.x` and `view->lastConfigure.y` during running B.Amp in `puglViewSize()` https://github.com/lv2/pugl/blob/609aed744bf205c992a10dab79109e3d6df1991d/src/x11_cairo.c#L23-L39 and both were 0 (as expected) in pugl but 100 in Carla (which is (estimated) the offset there).

If I do the same with B.Angr and force a full window redraw, I get -350, -130 which also looks like the negative offset in the screenshot mentioned above. I think these values are linked to the cause of the problem.

drobilla:
> NOT observed in old pugl builds (Jan 2021) I used before - Therefore I think it's pugl-related.

You mean currently, with the same current version of the same hosts, the old pugl code works properly?
Would those offsets happen to correspond to partial redisplays you're doing?
Does the issue still exist if you only ever do full redisplays of the entire view?

drobilla:
> and both were 0 (as expected) in pugl but 100 in Carla (which is (estimated) the offset there)

[...]

PR #99, opened 2023-02-18
Enable puglSetPositionHint prior puglRealize

sjaehn:
`puglSetPositionHint()` was without any effect if called prior `puglRealize()`. The main reason was that the "use the default position" condition in `getInitialFrame()` wasn't fulfilled and thus not used. The fallback for this was the window center algorithm which caused clipping in the majority of the LV2 hosts. And I just *started* to use the new position hints.

I made changes as discussed in #98. Win and mac edits were mainly copy/pasted from x11 but not tested.

I know you are currently working on this drafted branch. Feel free to take it, or get inspiration, or do something else with it.


drobilla:
Thanks, although the (usual) way I'm working on that branch isn't really compatible with literally merging this (or anything).  The current version incorporates most of the things here, if slightly differently.

Issue #42, opened 2021-01-26
UI size and resize miracle ;-)


sjaehn:
After our yesterdays discussion I did some experiments. Your idea was NOT to call ui_resize() from LV2UI and just let the host find the right size. Makes sense. But I observed that the plugin UI is then showed with its min size. In the most cases. Without any need (full HD screen). I don't know if this is a feature or a bug. And if it is somehow related to pugl (I've some doubts). Or more likely to the hosts and/or their libraries (suil?). But then I would keep calling ui_resize() following puglRealize().

I used and manipulated B.Oops and pugl in jalv.gtk3, jalv.qt5, Carla, ardour 6.x, and qtractor. The relevant plugin code contains (https://github.com/sjaehn/BOops/blob/master/src/BWidgets/Window.cpp#L47):
```
	puglSetDefaultSize (view_, getWidth (), getHeight ());
	puglSetMinSize (view_, getWidth () / 4.0, getHeight () / 4.0);
	//...
	puglSetViewHint(view_, PUGL_RESIZABLE, PUGL_TRUE);
	puglSetViewHint(view_, PUGL_IGNORE_KEY_REPEAT, PUGL_TRUE);
	//...
	puglSetBackend(view_, puglCairoBackend());
	//...
	puglRealize (view_);
	puglShow (view_);
```
with getWidth() = 1240, and getHeight() = 608.

As I told you yesterday, all plugin UIs are shown in a largely reduced size in all five conditions. In the min size in jalv, Carla, and Ardour. A bit larger (640x480) in qtractor. But puglRealize calls XCreateWindow() with 1240x608. XGetWindowAttributes() confirmed the window size of 1240x608.

And there are NO PUGL_CONFIGURE events emitted in jalv.qt5. And not in qtractor. As @ventosus mentioned, the only way to inform the plugin UI about a resize in qtractor is via the ui:resize extension. I can confirm this. This works now :-).

I added a stderr output in the event function if a PUGL_CONFIGURE event is emitted ("configure ", window.x, window.y, window.width, window.height). Interesting output, just after puglRealize():

```
(jalv.gtk3:124693): Gtk-WARNING **: 13:24:15.244: SuilX11Wrapper 0x556f7720dba0 reported min size 152 and natural size 0 in get_preferred_height(); natural size must be >= min size

[...]

PR #100, opened 2023-02-18
Fix creation zero-sized uis in jalv, ardour

sjaehn:
Me again ;-). Now I found a solution for the 2 years old issue #42 which is closed but not fixed yet.

Problem description:
Pugl-based Plugin UIs are realized in jalv and ardour with size zero (or maybe (1, 1) - I'm not sure) if no `ui_resize()` (provided as LV2 feature) is called instantly. Even if `puglSetSize()` or `puglSetSizeHint()` is called prior `puglRealize()`. jalv addidtionally prints error messages on the screen. To visualize the problem (jalv with B.Amp at 100x100 (but not displayed) is in top left):
Bildschirmfoto von 2023-02-18 21-46-22

`ui:resize` became deprecated as requested by faktx (AFAIK, didn't check). The *manual* resize mechanism (dragging) without resize as feature or interface fully works in jalv, ardour, reaper, zrythm, and carla. This was also one of the reasons to deprecate. Only the zero size-intitialized UIs in jalv and ardour stopped me to remove `resize` from my code, yet.

Solution:
I saw that @brummer10 uses `XSetWMNormalHints()` instead of `XSetNormalHints()`. 

Tests:
This works with all hosts tested (jalv, ardour, reaper, zrythm, carla, qtractor) without any error message. Also my standalone test apps (like B.Widgets widgetgallery) work as expected. 

Note:
Calling `puglSetSize()` or `puglSetSizeHint()` *after*  `puglRealize()` still produces different host-dependent results. Only to keep in mind, no TODO for now.

### DIFF
--- a/src/x11.c
+++ b/src/x11.c
@@ -419,7 +419,7 @@ updateSizeHints(const PuglView* const view)
     }
   }
 
-  XSetNormalHints(display, view->impl->win, &sizeHints);
+  XSetWMNormalHints(display, view->impl->win, &sizeHints);
   return PUGL_SUCCESS;
 }
 


drobilla:
I don't really understand most of what you're saying there, to be honest, but it seems reasonable enough to me.  Merged as 9b5a087, thanks.

PR #102, opened 2023-03-25
Fix puglStopTimer() X11, amend test_timer.c with according tests

7890:


drobilla:
Merged as 6b31fb5, thanks.

I rewrote the test to also check that multiple timers with different frequencies actually work properly: 0fac728

Issue #105, opened 2023-10-03
Cairo on MacOS draws nothing


falkTX:
Seems cairo on macOS has been broken for a long time, I thought it was me doing something wrong on DPF side but I just built the pugl cairo examples and still end up with an empty window.

Running macOS 12.6.1, cairo installed via homebrew. Using f6f0ecfa0aa9872a582e22f2b839d500105aafa1

Build log:

```
% meson build
The Meson build system
Version: 1.1.1
Source dir: /Users/falktx/Source/DISTRHO/DPF/dgl/src/pugl-upstream
Build dir: /Users/falktx/Source/DISTRHO/DPF/dgl/src/pugl-upstream/build
Build type: native build
Project name: pugl
Project version: 0.5.3
C compiler for the host machine: ccache cc (clang 14.0.0 "Apple clang version 14.0.0 (clang-1400.0.29.202)")
C linker for the host machine: cc ld64 820.1
Host machine cpu family: x86_64
Host machine cpu: x86_64
C++ compiler for the host machine: ccache c++ (clang 14.0.0 "Apple clang version 14.0.0 (clang-1400.0.29.202)")
C++ linker for the host machine: c++ ld64 820.1
Objective-C compiler for the host machine: ccache cc (clang 14.0.0)
Objective-C linker for the host machine: cc ld64 820.1
Objective-C++ compiler for the host machine: ccache c++ (clang 14.0.0)
Objective-C++ linker for the host machine: c++ ld64 820.1
Library dl found: YES
Run-time dependency appleframeworks found: YES (foundation, corevideo)
Library m found: YES
Found pkg-config: /usr/local/bin/pkg-config (0.29.2)
Run-time dependency gl found: YES
Run-time dependency cairo found: YES 1.16.0
Compiler for C supports arguments -Wno-documentation: YES 
Compiler for C supports arguments -Wno-documentation-unknown-command: YES 
Compiler for C supports arguments -Wno-reserved-identifier: YES 
Run-time dependency vulkan found: NO (tried pkgconfig and system)
Compiler for Objective-C++ supports arguments -Wno-c++98-compat: YES 
Compiler for Objective-C++ supports arguments -Wno-c++98-compat-pedantic: YES 
Compiler for Objective-C++ supports arguments -Wno-deprecated-declarations: YES 
Compiler for Objective-C++ supports arguments -Wno-direct-ivar-access: YES 
Configuring Info.plist using configuration
Configuring Info.plist using configuration
[...]

PR #106, opened 2023-10-03
Fix cairo-quartz context translate call

falkTX:
Fixes the issue reported in #105

### DIFF
--- a/src/mac_cairo.m
+++ b/src/mac_cairo.m
@@ -96,7 +96,7 @@ puglMacCairoEnter(PuglView* view, const PuglExposeEvent* expose)
   const CGSize sizePt = CGContextConvertSizeToUserSpace(context, sizePx);
 
   // Convert coordinates to standard Cairo space
-  CGContextTranslateCTM(context, 0.0, -sizePt.height);
+  CGContextTranslateCTM(context, 0.0, sizePt.height);
   CGContextScaleCTM(context, scale, -scale);
 
   drawView->surface = cairo_quartz_surface_create_for_cg_context(


drobilla:
This breaks the demo application for me.  Before this change, everything displays fine, after this, nothing is displayed.

src/mac_cairo.m
```diff
@@ -96,7 +96,7 @@ - (void)drawRect:(NSRect)rect
   const CGSize sizePt = CGContextConvertSizeToUserSpace(context, sizePx);
 
   // Convert coordinates to standard Cairo space
-  CGContextTranslateCTM(context, 0.0, -sizePt.height);
+  CGContextTranslateCTM(context, 0.0, sizePt.height);
   CGContextScaleCTM(context, scale, -scale);
```
mfisher31, line 100: welp.... shoulda looked at this commit first :)

mfisher31: wonder if the -height is only needed for OSX v12.x.x.  Wouldn't be great to add an ifdef for that, but could it be?

mfisher31: Wonder if these two lines of Y-axis flipping should be a `once` kind of deal.   cache a bool "initialized" member in the private view and call the flip one time?  maybe it's just flipping itself an odd amount of times when using mac SDK vX.Y.Z... just a thought.

mfisher31: note to all: I tested out the "once" idea.  no go on that.

drobilla: It was a good idea, but yeah, I think if that was happening a whole lot of things would be much more broken everywhere because this is called when the context is entered.

Issue #104, opened 2023-09-26
Adding pointer grab support


swesterfeld:
In some cases my SpectMorph plugin has the problem that the mouse pointer hits the border of the screen while  the user would still want to drag further. For instance SpectMorph has sliders which support SHIFT+drag to do fine adjustment. In this case after the mouse click the mouse is moving much faster than the slider itself.

The best fix I can think of is for these cases grab & hide the mouse pointer so that it will not hit the border of the screen no matter how far the user drags the mouse from its original click point. I wonder if this would be a feature you would want to see added to pugl. If so, I can try to provide an API suggestion and at least an implementation for X11 as PR.

drobilla:
Sure.  I'm not sure about how the details should work, probably should check out what other things do (presumably it gets used a lot in glfw since the gamer types certainly need it), but it seems like something that should be there anyway.  Moreso than some things I have reluctantly accepted, really 😄 

A preliminary implementation for X11 would be great, thanks.  I can hopefully make the same thing work on Windows and MacOS easily enough.

PR #107, opened 2023-10-08
Pointer lock API / example for X11

swesterfeld:
As discussed in issue #104 it would be great if pugl supported grabbing/locking the mouse cursor. Here is my attempt to support this in pugl. I've only implemented the X11 version. The API has been inspired by the Qt API. They don't really have an extra "lock/grab the pointer" functionality. Instead, you can build your own based on two primitives:

* hide the mouse cursor
* set the mouse cursor position

So this branch adds these two primitives and then a pugl pointer lock example which has a simple control that will "lock" the pointer while pressed, supporting normal and fine adjustment (which would hit the border of the screen without pointer lock).

From an API point of view the difference between Qt and my suggestion is that in the Qt API you set the cursor position with global coordinates, whereas here this is using local coordinates relative to the view. I don't know if there are good reasons to use global coordinates in pugl.

For comparision, here is how you can implement a pointer lock using Qt:
```
#include <QApplication>
#include <QWidget>
#include <QCursor>
#include <QMouseEvent>

class CustomWidget : public QWidget
{
public:
  CustomWidget(QWidget *parent = nullptr) : QWidget(parent)
  {
  }

protected:
  QPointF orig_pos;
  QPointF acc_pos;
  bool dragging = false;

  void
  mousePressEvent (QMouseEvent *event) override
  {
    if (event->button() == Qt::LeftButton)
      {
        setCursor(Qt::BlankCursor);
        orig_pos = mapToGlobal (event->pos());
        acc_pos = QPointF (0, 0);
        dragging = true;
      }
  }

  void
  mouseMoveEvent (QMouseEvent *event) override
  {
    if (dragging)
      {
        auto global_pos = mapToGlobal (event->pos());
        if (global_pos != orig_pos)
          {
            auto delta_pos = global_pos - orig_pos;
            acc_pos += delta_pos;
            printf ("%f %f\n", acc_pos.x(), acc_pos.y());
            QCursor::setPos (orig_pos.x(), orig_pos.y());
          }
      }
  }

  void
  mouseReleaseEvent (QMouseEvent *event) override
  {
    if (event->button() == Qt::LeftButton)
      {
        QCursor::setPos (orig_pos.x(), orig_pos.y());
        setCursor (Qt::ArrowCursor);
        dragging = false;
      }
  }
};

int main(int argc, char *argv[])
{
  QApplication app(argc, argv);

  CustomWidget widget;
  widget.setGeometry (100, 100, 400, 300);
  widget.show();

  return app.exec();
}
```

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -20,6 +20,7 @@ gl_examples = [
 cairo_examples = [
   'pugl_cairo_demo.c',
   'pugl_management_demo.c',
+  'pugl_pointer_lock_demo.c',
 ]
 
 vulkan_examples = [
--- a/examples/pugl_pointer_lock_demo.c
+++ b/examples/pugl_pointer_lock_demo.c
@@ -1,0 +1,205 @@
+// Copyright 2012-2020 David Robillard <d@drobilla.net>
+// Copyright 2023 Stefan Westerfeld <stefan@space.twc.de>
+// SPDX-License-Identifier: ISC
+
+#include "demo_utils.h"
+#include "test/test_utils.h"
+
+#include "pugl/cairo.h"
+#include "pugl/pugl.h"
+
+#include <cairo.h>
+
+#include <math.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+
+typedef struct {
+  PuglWorld*      world;
+  PuglTestOptions opts;
+  double          mouseClickX;
+  double          mouseClickY;
+  double          value;
+  int             quit;
+  bool            entered;
+  bool            mouseDown;
+} PuglTestApp;
+
+typedef struct {
+  double x;
+  double y;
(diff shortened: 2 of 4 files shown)


drobilla:
Thanks.  Sorry for the massive delay, I'll try to find some time to implement this on the other platforms.

drobilla:
I don't think `puglSetCursorPos` should exist, programs should never move the mouse cursor and providing such a function invites abuse and makes bugs (especially cross-platform ones) likely.  Pugl should only support hiding the cursor and getting "virtual" movement events, and the API should reflect that.  Then, the implementation can handle this internally, and applications can rely on it working properly everywhere.

Probably the approach itself is about the same, but the pointer warping (if it's actually necessary on a given platform) needs to be handled internally.

PR #110, opened 2023-10-21
tests: print modifiers for key events

falkTX:
Adds modifiers to the prints during key events, needed in order to verify the ticket at #111


drobilla:
Handy.  Merged as f2be8e2, thanks.

Issue #111, opened 2023-10-21
Inconsistent modifier key behaviour across platforms


falkTX:
When using a modifier flag like "shift" or "control" we have inconsistent behaviour across platforms. Mainly on X11 vs macOS and win32, that only register the modifier key after another key press.

With #110 in place, we can see the behaviour of the shift-key modifier:

```
Event: Key press   code  50 key  U+E051 (SHIFT_L) Modifiers:
Event: Update
Event: Key release code  50 key  U+E051 (SHIFT_L) Modifiers: Shift
Event: Update
```

Notice on how the first event the shift key is pressed, but it has no shift modifier in place yet. On key release the modifier is there.

Now on windows the following behaviour occurs:

```
Event: Key press   code  42 key  U+E051 (SHIFT_L) Modifiers: Shift
Event: Update
Event: Key release code  42 key  U+E051 (SHIFT_L) Modifiers:
Event: Update
```

Notice how the shift modifier is in place for the initial key press, and not present on the release event.
Same happens on macOS, though I dont have the logs to confirm it right now.

Which behaviour is correct is sorta debatable, the main issue is the inconsistency across platforms.

drobilla:
Can we just mask the corresponding ones out entirely?  It seems like nonsense to ask whether the shift key is shift modified anyway, so an easy route to consistency is for that to never happen as far as a pugl client is concerned.

falkTX:
@dromer was the one that initially found the issue, so maybe he has a specific usecase for it

dromer:
It's a very normal use-case: use it as a modifier key.

The faulty behavior is easily spotted with the https://github.com/DISTRHO/dear-plugins ImGuiDemo - compared to the default desktop ImGuiDemo.

drobilla:
There's too many layers there I'm not familiar with to bother wading through, but it seems like anything that breaks depending on the modifier bit being set or not for a modifier key that's being pressed/release is probably doing something... questionable.

Doesn't really matter though, it should be consistently _something_, and the platform behaviours seems arbitrary and meaningless, so the easiest thing to do is just strip them out, always.  It makes no sense to ask about whether the shift key _that's currently being pressed/released_ is shift modified, so that's a very simple consistent thing Pugl can guarantee, and it's easy to implement for any platform.

So, fixed in 6c2a219.  Tested on X11, Windows, and MacOS.  Also fixed left vs right modifier detection on Windows, although that still doesn't work on MacOS.

PR #112, opened 2023-10-21
Make modifier keys behaviour consistent across platforms

falkTX:
Fixes the modifier key state not being consistent across platforms, reported as #111

Feel free to change, this is more of a proposal than a full fix, as we likely should be handling the L vs R keys separately.


drobilla:
Inconsistency here fixed more, let's say "aggressively consistently", in 6c2a219.

PR #114, opened 2023-10-21
x11: Initialize arrays in C++ friendly way

falkTX:
A little minor thing I noticed while debugging something else.
Weirdly I never got a compiler warning from these.


drobilla:
If everyone could just agree that {} makes sense here, 10 years ago, that'd be greeeeeat.

Merged as 0fed0b5, thanks.

PR #115, opened 2023-10-21
x11: fix keyboard input with control modifier active

falkTX:
In X11, using Ubuntu 22.04 KDE Plasma desktop, I could not get proper keyboard events while the control key was held down.

I added a print right after the x11 utf8 key decode, like so:
```
  printf("x11 special:%d ufound:%d key:%d:'%c' %02x:%02x:'%s'\n",
           special, ufound, event->key.key, event->key.key, ustr[0], ustr[1], ustr);
```

And this was the print out for Ctrl+C:

```
x11 special:0 ufound:1 key:3:'' 03:00:''
```

After the fix from this PR, the following is printed, and yes it then works as expected:

```
x11 special:0 ufound:1 key:99:'c' 63:00:'c'
```

### DIFF
--- a/src/x11.c
+++ b/src/x11.c
@@ -880,8 +880,8 @@ translateKey(PuglView* const view, XEvent* const xevent, PuglEvent* const event)
 
   event->key.keycode = xevent->xkey.keycode;
 
-  // Mask off the shift bit to get the lowercase "main" symbol
-  xevent->xkey.state = xevent->xkey.state & ~(unsigned)ShiftMask;
+  // Mask off the control and shift bits to get the lowercase "main" symbol
+  xevent->xkey.state = xevent->xkey.state & ~(unsigned)(ControlMask|ShiftMask);
 
   // Lookup unshifted key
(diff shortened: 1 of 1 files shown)


drobilla:
This fixes (arguably, I guess) the key, but there's still text entry of a control character.

drobilla:
On X11, that's just what the lookup stuff does.  I don't know if all such events should just be filtered, or if this breaks whatever input method, or...

falkTX:
I checked with other modifiers, Alt and Meta still get the key properly, only Control modifier made it lose the regular keys.

drobilla:
Yeah, it's because the control key _is_ a set bit, at least, uh "traditionally".  In this case, I guess they should "always" (?) be filtered out, and it seems weird that the X string lookup stuff does this, but whatever I guess, that's a separate issue.  Seems obvious the PuglKeyPressEvent::key shouldn't be shifted/"controlled" here regardless.

Merged as a7803c9, thanks.

PR #117, opened 2023-10-22
X11: Only call XSetTransientForHint for valid windows

falkTX:
While testing a few different WMs, noticed this warning from metacity:

```
(metacity:586692): metacity-WARNING **: 15:11:43.796: Invalid WM_TRANSIENT_FOR window 0x0 specified for 0x5c00002 (Cardinal).
```

PR fixes it by ensuring the transient parent is valid (non-zero).


drobilla:
Merged as 5d8b3d3, thanks.

PR #127, opened 2024-12-12
quick and dirty hack to generate text events for numpad keys

MartinPreinfalk:
As mentioned in https://github.com/lv2/pugl/issues/121 it would be nice to have some text events for the numpad keys.

 I therefore have prepared a quick and dirty hack to showcase it for Linux X11. I do not recommend to use this in production. It needs to be implemented "properly" and for all the platforms.

I do this in my spare time and I am definitely no expert in X11 coding, so I don't know if I did this properly, as said it's just a hack (you have been warned :-))

I cannot provide solutions for the other platforms, I only have Linux machines.

### DIFF
--- a/src/x11.c
+++ b/src/x11.c
@@ -890,14 +890,14 @@ translateKey(PuglView* const view, XEvent* const xevent, PuglEvent* const event)
   KeySym        sym     = 0;
   const int     ufound  = XLookupString(&xevent->xkey, ustr, 8, &sym, NULL);
   const PuglKey special = keySymToSpecial(sym);
-  if (special) {
+  if (special && ufound <= 0) {
     event->key.state = puglFilterMods(event->key.state, special);
     event->key.key   = special;
   } else if (ufound > 0) {
     event->key.key = (PuglKey)puglDecodeUTF8((const uint8_t*)ustr);
   }
 
-  if (xevent->type == KeyPress && !filter && !special && view->impl->xic) {
+  if (xevent->type == KeyPress && !filter && !(special && ufound <= 0) && view->impl->xic) {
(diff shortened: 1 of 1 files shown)


drobilla:
Thanks.  A quick and dirty hack doesn't get us very far on this one, though.  Last I tried, keypad num lock is considerably more of a pain to get working properly than you'd expect.

I don't remember how the other platforms went but apparently I didn't keep that branch around.

drobilla:
My mistake, I was misremembering the evolution of the keypad support (which was mostly implemented in a347ffd2f45feb4e180dd470046436f51591bd91).  I tested things (with virtual keyboards) on all three platforms, currently:

* On X11, as you say, keypad press and release events work but there's no text entry events for symbols.

* On Windows, the standard PC keyboard keypad (with both "navigation" and "number" layers toggled by Num Lock) works, with text entry for symbols (mainly numbers) when Num Lock is on.

* On MacOS, Apple-style extended keyboard keypads (with a single-layer) work, with text entry for symbols.  I don't know (or really care) how full PC keyboards work.

I suppose it's not surprising that I don't use a full keyboard on my main workstation :)  It's easy to test with a virtual keyboard, though.

drobilla:
Your patch seems half-right: the first change breaks the special key press/release events, but the second one fixes the issue.  I'm not sure if `!(special && ufound <= 0)` there is exactly right... X11 input is tricky, it might be a coincidence that "not a special key with no looked-up string" matches the keys we want to send text events for.  As far as I can tell, it works with X.org for every keypad key on both layers of a standard English IBM PC keyboard, anyway.

I merged the second change as e2f89df, thanks.

PR #129, opened 2025-02-08
Fix macOS embed UI resize from host

falkTX:
Initially discovered in https://github.com/trummerschlunk/PodcastPlugins/issues/22

Ardour does not currently support fixed-size LV2 UIs, it always allows resizing.
This works fine on Linux/X11 but was causing some odd issues on macos, related to the ticket above.

When resizing through the plugin ui (and Ardour then adjust to it), everything seems well.
After some tests I noticed that the opengl draw (ns)view did not have the correct size applied when resized from the host, even though the main wrapper view did.

Maybe we assumed that `[impl->wrapperView setAutoresizesSubviews:YES];` would be enough, but it is clearly not, leading to this issue.

Fix is possible by forcing the draw view to be resized, only done for the cases where we dont set size just before

### DIFF
--- a/src/mac.m
+++ b/src/mac.m
@@ -188,13 +188,18 @@ getCurrentViewStyleFlags(PuglView* const view)
 }
 
 static PuglStatus
-dispatchCurrentChildViewConfiguration(PuglView* const view)
+dispatchCurrentChildViewConfiguration(PuglView* const view, bool drawViewResize)
 {
   const NSRect framePt = [view->impl->wrapperView frame];
   const NSRect framePx = nsRectFromPoints(view, framePt);
 
   if (view->stage < PUGL_VIEW_STAGE_REALIZED) {
     return PUGL_SUCCESS;
+  }
+
+  if (drawViewResize) {
+    const NSSize sizePt = [view->impl->drawView convertSizeFromBacking:framePx.size];
+    [view->impl->drawView setFrameSize:sizePt];
   }
 
   const PuglConfigureEvent ev = {
@@ -317,7 +322,7 @@ dispatchCurrentChildViewConfiguration(PuglView* const view)
     if (puglview->impl->window) {
       [puglview->impl->window dispatchCurrentConfiguration];
     } else {
-      dispatchCurrentChildViewConfiguration(puglview);
+      dispatchCurrentChildViewConfiguration(puglview, true);
     }
     reshaped = false;
   }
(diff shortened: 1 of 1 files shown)


falkTX:
To be clear, because that other ticket from PodcastPlugins has a few more things involved, the issue on pugl side is this:

image

Ardour resizes the 1st wrapper view, but the draw view inside it would not be updated and thus rendered incorrectly.

src/mac.m
```diff
@@ -1767,7 +1772,7 @@ - (void)windowDidExitFullScreen:(NSNotification*)notification
 
   [impl->wrapperView setFrame:framePt];
   [impl->drawView setFrame:sizePt];
```
drobilla, line 1774: Now I'm confused, calling setFrame (not `setFrameSize`) on an NSView here, not an NSWindow, which doesn't even seem to be a method.  Although I'm just about to nuke `puglSetFrame` entirely (combining size and position is the devil, both for Wayland, and in general due to embedding and scaling bugs), so I guess this particular case doesn't matter anyway.

src/mac.m
```diff
@@ -1806,7 +1811,7 @@ - (void)windowDidExitFullScreen:(NSNotification*)notification
   [impl->drawView setFrameOrigin:drawPt.origin];
 
   // Dispatch new configuration
-  return dispatchCurrentChildViewConfiguration(view);
+  return dispatchCurrentChildViewConfiguration(view, false);
```
drobilla, line 1814: Not so much false because "we [...] set size just before", but because this function must not change the size at all.

src/mac.m
```diff
@@ -197,6 +197,11 @@
     return PUGL_SUCCESS;
   }
 
+  if (drawViewResize) {
```
drobilla, line 200: I don't think this is an appropriate place to be messing with the views, before this function did just what it said on the tin, now it's `dispatchCurrentChildViewConfigurationExceptSometimesMaybeResizeTheDrawView`.

drobilla (commented):
Thanks, although, bad timing, I'm gutting some of the problematic size/pos stuff as we speak.  I left some review comments as notes, but don't waste any time trying to address them in the code.

src/mac.m
falkTX: nsviews do have setFrame, as they have position and size. subviews can be offset from their parent

drobilla:
Seems dropping support for noUserResize is forcing bugs to be discovered and fixed now.  In retrospect everyone should have done that ages ago, if the experience with size and position stuff has taught us anything, it's that workarounds only make things worse.

src/mac.m
drobilla: Come to think of it, at a higher level, doing this in `dispatchExpose` seems like a kludge.  This should have happened earlier, although I'm not sure where off the top of my head.

PR #130, opened 2025-03-01
X11: fix BadWindow for embedded windows when the parent is destroyed

Quant1um:
This pull request fixes a crash in plugin context in several DAWs (mainly Bitwig) where the host destroys the parent window before calling the GUI destroy function (`clap_plugin_gui::destroy` for CLAP or `IPlugView::removed` in VST3). 

Calling `puglFreeView` there would result in a crash due to calling `XDestroyWindow` for a non-existent (already destroyed) window. This bug is possible in any situation in which a parent window is destroyed before we can call `puglFreeView`.

Fortunately the fix is straightforward, X11 provides a way to know when our window was destroyed by listening to `DestroyNotify` event.

### DIFF
--- a/src/x11.c
+++ b/src/x11.c
@@ -1190,6 +1190,9 @@ translateEvent(PuglView* const view, XEvent xevent)
     view->impl->mapped = false;
     event              = makeConfigureEvent(view);
     break;
+  case DestroyNotify:
+    view->impl->win = None;
+    break;
   case ConfigureNotify:
     event                  = makeConfigureEvent(view);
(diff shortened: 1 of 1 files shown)


drobilla:
> where the host destroys the parent window before calling the GUI destroy function

That doesn't sound like something that can be reliably worked around, there's countless places where `win` being NULL will explode.  That said, since it explodes anyway, and this can't break anything if the window _hasn't_ been destroyed, I'll take your word for it.

Merged as ff2634bb, thanks.

PR #138, opened 2026-03-01
Adjust PUGL_KEY_F11, PUGL_KEY_F12

7890:
mac.m depends on contiguous F-key numbering:
if (sym >= NSF1FunctionKey && sym <= NSF12FunctionKey) {
      return (PuglKey)(PUGL_KEY_F1 + (sym - NSF1FunctionKey));
}
I have tested this with Pugl Event Printer (try F12).


drobilla:
The caps lock handling isn't quite right, it sends key release events when caps lock is _disabled_ (on a press when it's enabled). I don't know if it's possible to handle caps lock release on MacOS without jumping through hoops.

Merged as 6fff46cf 8d3cac70, thanks.

7890:
Thanks @drobilla. I'm preparing an in-depth test for all keyboard input to spot oddities across platforms.

